PR #1, opened 2019-06-12
Globally replace 'Your Teaching Potential' with 'Your details'

duncanjbrown:
Content update from Laura Tennant

### DIFF
--- a/app/views/profile/motivation.njk
+++ b/app/views/profile/motivation.njk
@@ -9,7 +9,7 @@
       text: "Home",
       href: "/dashboard"
     }, {
-      text: "Your teaching potential",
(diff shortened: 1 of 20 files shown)


app/views/profile/motivation.njk
```diff
@@ -22,7 +22,7 @@
   <ul class="govuk-list govuk-list--bullet">
     <li>convince providers of your passion for teaching</li>
     <li>explain why you want to work with children</li>
-    <li>make the case for your teaching potential</li>
+    <li>make the case for Your details</li>
```
kjdchapman:
```suggestion
    <li>make the case for your teaching potential</li>
```

PR #71, opened 2019-08-09
Full course picker

fofr:
- Add a list of all providers and courses for autocompletes
- Create journey to pick a course
- Copy course to application on completion
- Move filters out of server.js


package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "prestart": "npm install",
     "start": "node start.js",
+    "s": "node start.js",
```
paulrobertlloyd: Smart!

app/views/course/provider.njk
```diff
@@ -5,20 +5,27 @@
...
-      text: 'Provider name'
-    },
-    autocomplete: "name",
-    id: "candidate-name",
-    name: "candidate[name]",
-    value: data["candidate"]["name"]
-  }) }}
+  <div class="govuk-form-group">
+    <label for="courses-{{ courseId }}-provider" class="govuk-label">Provider name</label>
+    <div id="provider-autocomplete-containter"></div>
```
paulrobertlloyd: I take it you chose to use an empty `div` rather than enhance a `select` due to the amount of data involved? I know it’s only a prototype at this stage, but what would the non-JS fallback be if this design were to make it into the live service? Just enhance an `input type="text"` and use server-side validation?

fofr: We can probably enhance a text input in the way that Find does it.

paulrobertlloyd: Maybe we should do that here then (or as close as we can get to that). Just wary of inaccessible code slipping into the prototype (and the possibility of this then causing issues in usability sessions).

PR #80, opened 2019-08-13
Add course content

laurahermionetennant92:
Add course content

### DIFF
--- a/app/views/course/found.njk
+++ b/app/views/course/found.njk
@@ -1,6 +1,6 @@
 {% extends "_form.njk" %}
 
-{% set title = "Have you found a course to apply to?" %}
+{% set title = "Have you chosen a course to apply to?" %}
 {% set formaction = paths.current %}
 {% block pageNavigation %}{{ govukBackLink({ href: paths.back }) }}{% endblock %}
 
@@ -8,7 +8,7 @@
   {{ govukRadios({
     items: [{
       value: "know",
-      text: "Yes, I know who I want to apply to"
+      text: "Yes, I know where I want to apply"
     }, {
       value: "need",
       text: "No, I need to find a course"
--- a/app/views/course/pick.njk
+++ b/app/views/course/pick.njk
@@ -7,7 +7,7 @@
 
 {% block primary %}
   <div class="govuk-form-group">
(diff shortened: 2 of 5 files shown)


app/views/course/pick.njk
```diff
@@ -7,7 +7,7 @@
 
 {% block primary %}
   <div class="govuk-form-group">
-    <label for="courses-{{ courseId }}-course" class="govuk-label">Course name and code</label>
+    <label for="courses-{{ courseId }}-course" class="govuk-label">Enter course name and code</label>
```
fofr: I don't think we should be adding "Enter" here.

app/views/course/found.njk
```diff
@@ -1,14 +1,14 @@
 {% extends "_form.njk" %}
 
-{% set title = "Have you found a course to apply to?" %}
+{% set title = "Have you chosen a course to apply to?" %}
```
fofr: Nice :100:

app/views/course/provider.njk
```diff
@@ -1,12 +1,12 @@
...
 
-{% set title = "Which provider are you applying to?" %}
+{% set title = "Which training provider are you applying to?" %}
 {% set formaction = paths.next %}
 {% block pageNavigation %}{{ govukBackLink({ href: paths.back }) }}{% endblock %}
 
 {% block primary %}
   <div class="govuk-form-group">
-    <label for="courses-{{ courseId }}-provider" class="govuk-label">Provider name</label>
+    <label for="courses-{{ courseId }}-provider" class="govuk-label">Enter training provider name</label>
```
fofr:
```suggestion
    <label for="courses-{{ courseId }}-provider" class="govuk-label">Training provider</label>
```

app/views/course/found.njk
fofr: Although, thinking about this some more in context – the purpose of these screens is to choose the provider and course. I wonder if this might be interpreted as "not yet, that's what I'm doing".

The thing to communicate is essentially "do you know what course you're applying to"

app/views/course/pick.njk
fofr: Spoke to @laurahermionetennant92 about this, we'll keep the label the same but add some help text to describe what a user must do.

PR #82, opened 2019-08-13
Replace breadcrumbs with Back links

fofr:
- Adds Back links to the easy use cases
- Removes add/edit logic, show the same page whether editing or not
- Fix default course choices

### DIFF
--- a/app/views/application/contact-details/address.njk
+++ b/app/views/application/contact-details/address.njk
@@ -1,9 +1,7 @@
 {% extends "_form.njk" %}
+
 {% set parent = "Contact details" %}
 {% set title = "What is your address?" %}
-
-{% block pageNavigation %}
-{% endblock %}
 
 {% block primary %}
   {{ govukInput({
--- a/app/views/application/contact-details/address.njk
+++ b/app/views/application/contact-details/address.njk
@@ -10,9 +10,6 @@
 {% endif %}
 
 {% block pageNavigation %}
-  {{ govukBreadcrumbs({
-    items: applicationBreadcrumbItems
-  }) }}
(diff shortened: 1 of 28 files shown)


app/views/application/contact-details/address.njk
```diff
@@ -10,9 +10,6 @@
 {% endif %}
 
 {% block pageNavigation %}
-  {{ govukBreadcrumbs({
-    items: applicationBreadcrumbItems
-  }) }}
 {% endblock %}
```
paulrobertlloyd: Empty `pageNavigation` block?

app/views/application/contact-details/select-address.njk
```diff
@@ -1,5 +1,4 @@
 {% extends "_form.njk" %}
-
```
paulrobertlloyd: Stylistically, have been separating `extends` from the `set` statements that follow. Can we keep that?

app/views/application/contact-details/address.njk
fofr: Something needs to go in here eventually. I just removed this so I could safely delete `applicationBreadcrumbItems`

paulrobertlloyd:
```suggestion
  {# TODO: Add back link #}
{% endblock %}
```

app/views/application/contact-details/select-address.njk
fofr: Updated

app/views/application/contact-details/address.njk
fofr: I've deleted it

PR #85, opened 2019-08-13
GCSE branching

paulrobertlloyd:
* Moves GCSEs out from qualifications
* Uses question flow similar to that for degrees: Type → (NARIC) → Grade → Year.
* What’s left of qualifications is now the “Is there anything you’d like to tell us about your qualifications (optional)?” question, poorly labelled “Anything else?” on the application screen. This clearly needs work!
* Need to add back logic to ask for science GCSE if primary application
* There will be bugs, but this covers most of the mechanics.


app/views/application/gcse/year.njk
```diff
@@ -0,0 +1,27 @@
...
+{% set formaction = referrer or applicationPath %}
+
+{% set parent = (gcseId | capitalize) + " GCSE or equivalent" %}
+{% if applicationValue(["gcse", gcseId, "grade"]) == "Pending" %}
+  {% set pending = true %}
+{% endif %}
+{% if pending %}
+  {% set title = "When was your qualification awarded?" %}
+{% else %}
+  {% set title = "When will you get your qualification?" %}
```
fofr: These two titles are the wrong way round.

fofr:
I think we need a definite pattern to each section:
[ Application ] -> [ Questions ] -> [ Questions ] -> [ Answers ] -> [ Application ]
Then when coming back to edit:
[ Application ] -> [ Answers ] -> [ Questions ] -> [ Answers ] -> [ Application ]

So each section linked to from the application takes you to:
a) The first question if unanswered
b) A summary of everything already entered

fofr:
Screen Shot 2019-08-13 at 20 48 17

This could be interpreted as someone with a pending qualification. Which means they'd miss out on the later questions.

PR #86, opened 2019-08-14
Increase text size on application index and when reviewing

fofr:
localhost_3000_application_12345_review
localhost_3000_application_12345


paulrobertlloyd:
Overall, I like this. The headings on the application don’t match those on the review page anymore (27px vs 24px), but can live with that (we need to er, review, the review page anyway).

The only other thing… course selection type sizing now seems to be out of step with the sizing used elsewhere; can those go one bigger?

PR #106, opened 2019-08-19
Content tweaks

laurahermionetennant92:
Content tweaks


app/views/application/degree/index.njk
```diff
@@ -1,6 +1,6 @@
 {% extends "_form.njk" %}
 
-{% set title = "Degree" %}
+{% set title = "Add degree" %}
```
fofr: I think this needs some logic around it @paulrobertlloyd

paulrobertlloyd: Done!

PR #159, opened 2019-09-12
Contextualize home link

paulrobertlloyd:


app/views/_layout.njk
```diff
@@ -56,10 +56,20 @@
     {% set signOutText = "Sign out" %}
   {% endif %}
 
+  {% if isSignedOut != true %}
```
fofr: `if not isSignedOut`?

PR #176, opened 2019-09-19
Re-align session data defaults with updated fields

benilovj:


app/data/session-data-defaults.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  account: {
+    email: 'janina.doe@example.com'
```
fofr: I think we need to mark this as `disabled_account` too.

benilovj: Fixed and force-pushed.

PR #179, opened 2019-09-19
Add feature flags

fofr:
Allow us to turn off:

- nationality
- international addresses
- address lookup
- international qualifications

### DIFF
--- a/app/views/application/personal-details/index.njk
+++ b/app/views/application/personal-details/index.njk
@@ -76,28 +76,30 @@
     }]
   }) }}
 
-  {{ appAutocomplete({
-    label: {
-      text: "Nationality",
-      classes: "govuk-label--m"
-    },
-    items: nationalities
-  } | decorateApplicationAttributes(["candidate", "nationality"])) }}
-
-  {% set anotherNationality %}
+  {% if data.flags.nationality %}
     {{ appAutocomplete({
       label: {
-        text: "Second nationality"
+        text: "Nationality",
+        classes: "govuk-label--m"
       },
       items: nationalities
-    } | decorateApplicationAttributes(["candidate", "second-nationality"])) }}
-  {% endset -%}
+    } | decorateApplicationAttributes(["candidate", "nationality"])) }}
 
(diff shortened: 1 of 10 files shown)


app/views/application/personal-details/index.njk
```diff
@@ -76,28 +76,30 @@
...
+          text: "Second nationality"
+        },
+        items: nationalities
+      } | decorateApplicationAttributes(["candidate", "second-nationality"])) }}
+    {% endset -%}
+
+    {{ govukDetails({
+      summaryText: "Add another nationality",
+      html: anotherNationality,
+      open: ["candidate", "second-nationality"] | ifSetForApplication
```
paulrobertlloyd: What does `ifSetForApplication` do? I’d not spotted this before.

app/views/application/personal-details/index.njk
```diff
@@ -76,28 +76,30 @@
...
-  {{ appAutocomplete({
-    label: {
-      text: "Nationality",
-      classes: "govuk-label--m"
-    },
-    items: nationalities
-  } | decorateApplicationAttributes(["candidate", "nationality"])) }}
-
-  {% set anotherNationality %}
+  {% if data.flags.nationality %}
```
paulrobertlloyd: We will still want to ask for nationality in the MVP, but not do any branching branch based on this answer.

paulrobertlloyd:
Really impressed by how little needed to be changed to support these different variations!

fofr:
> For international address, if international_address is true, show show structured fields with internationalised labels (i.e. ’Zip or postal code’, ’State or region’, etc), rather than a textarea.
Is this already a feature?

app/views/application/personal-details/index.njk
fofr: I haven't changed it here, just whitespace changes.

It's probably redundant:
```nunjucksAppEnv.addFilter('ifSetForApplication', (sections) => {
      return !!getApplicationValue(sections)
    })```

PR #182, opened 2019-09-20
Deleting things, part 1 (adding remove links)

paulrobertlloyd:
- Adds a new Nunjucks component (`app-summary-card`). This has an optional header (that contains a title and optional action links) and whose body accepts a chunk of HTML.
- Uses this new component for review pages.
- Splits up review includes into individual item includes, and includes that include multiple of these items, (where applicable).

~~In the process of reviewing to make sure nothing is broken. So far, known issues include:~~
- ~~Can’t edit the data of a GCSE~~
- ~~May need to resolve conflicts with MVP changes on `master`~~


fofr:
From testing this end to end:

- After adding my first school exp I can’t add another it seems
- Personal statement review doesn’t show values (works fine in Review page, just not when clicking from Application to the section)
- No “Completed” section against personal statement either
- Maths and English GCSE show fine on the review page, but not on the application section
- If I put in a date in work history without a preceding 0 it gets InvalidDateTime showing (eg 1 instead of 01)

PR #191, opened 2019-09-23
Deleting data

### LINKED ISSUE #150: Delete degree
We should allow people to delete degrees, yes?

paulrobertlloyd:
- Links remove links on summary cards to remove page
- Adds a review page
- Adds `get` route for removing items, that shows a confirmation page
- Adds `post` route for removing items, that deletes item object from session data
- Fixes a few things in work history around gaps/jobs (titles and an incorrect links)

Fixes #150.

Screen Shot 2019-09-23 at 16 44 29

**NOTE: This PR does not include removing courses. That’s gunna be a bit more involved.**

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -16,5 +16,6 @@ router.all(['/application/:applicationId/course/:courseId', '/application/:appli
 require('./routes/application')(router)
 require('./routes/course')(router)
 require('./routes/email')(router)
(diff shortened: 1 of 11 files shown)


fofr:
💯 Lovely stuff

I think at some point we'll also need a "successfully removed" flash message on the page.

PR #227, opened 2019-10-15
Stub "Apply on UCAS" page in course flow

fofr:
- keep full list of providers so candidates can always find who they’re looking for
- if a provider who isn’t in pilot is picked, give a route to UCAS, maybe some guidance, maybe a link to the dual running page on GiT

Screen Shot 2019-10-15 at 15 09 51


fofr:
Screen Shot 2019-10-17 at 13 43 26

PR #234, opened 2019-10-16
Move courses into own section (and highlight that can choose multiple)

paulrobertlloyd:
Work in progress, requires review. Has the following bugs:

- ~~For the life of me, I can‘t get ‘I’ve completed this section’ checkbox to submit data; should update `data.applications[applicationId].completed.degree` to `true` but doesn’t.~~ I’m an idiot.
- The courses (now choices) flow was built without editing a choice afterwards in mind. I’ve added links to edit a location and a course, but:
  - Selecting a different location isn’t reflected when you return to overview page,
  - Attempting to change a course gives error `TypeError: Cannot read property '1' of null`.

I think it’s all to do with the creation of `temporaryChoice` when you make choice, but not needing that once a choice exists in the application data.

Questions:
- It’s possible we could go straight into adding first course when selecting this option from the index page, but I suspect @laurahermionetennant92 will welcome the opportunity to provide some content on this page regarding choices.
- Where in the order of sections do we want to show this section; first or last.

As I say, WIP…


fofr:
Bugs fixed.

PR #236, opened 2019-10-17
Move pilot eligibility questions to cover all routes in

paulrobertlloyd:
## What’s changed

- Moves pilot eligibility questions so that they appear after the ‘Start now’ action from the service landing page. This is so that we can shift candidates regardless of their route in.
- Interstitial from Find is now a simple choice between the respective Apply and UCAS URLs (if course/provider is opted in).

### Check you are eligible
`/account/eligibility`

Screenshot 2019-10-17 at 16 07 45

### Not eligible
`/account/ineligible`

Screenshot 2019-10-17 at 16 07 37

## Things to check
- The content on the ineligible page needs review. Is providing links to related content in a sidebar the right approach. Are those the right links?
- If you are eligible, is ‘Before you start’ still the right title for the following page?
- Should the page we show when choosing a non-opted on course align more closely with this design (i.e. with the sidebar links)?

## Improvements
- When we add carrying over course details from Find, we should also ensure that functionality is used on this flow so that we show the provider and course codes to use when applying via UCAS application.

### DIFF
--- a/app/assets/sass/components/_related.scss
+++ b/app/assets/sass/components/_related.scss
@@ -1,8 +1,21 @@
 .app-related {
-  border-top: govuk-spacing(1) solid govuk-colour("blue");
+  border-top: 2px solid govuk-colour("blue");
   padding-top: govuk-spacing(3);
 }
 
 .app-related + .app-related {
   border-top-color: transparent;
+}
+
+.app-related__body {
+  @include govuk-font(16);
+}
+
+.app-related__list {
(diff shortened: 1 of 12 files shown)


fofr:
My preference is for all content to be inline.
If it's important enough to be there, it shouldn't need to be put on the right.

paulrobertlloyd:
You might be right. My sense is that this content isn’t important, but is useful. It’s really context about the pilot. We can test, and see.

PR #242, opened 2019-10-22
Content save and continue

laurahermionetennant92:
I've done some save and continue content, but I think it should really go under the header (Course choices review). Couldn't figure out how to do that. This could work for all the review sections I think.
Screenshot 2019-10-22 at 09 16 34

### DIFF
--- a/app/views/application/choices/index.njk
+++ b/app/views/application/choices/index.njk
@@ -3,7 +3,7 @@
 {% set applicationPath = "/application/" + applicationId %}
 {% set formaction = applicationPath %}
 {% set choices = applicationValue("choices") %}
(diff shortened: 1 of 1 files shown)


app/views/application/choices/index.njk
```diff
@@ -3,7 +3,7 @@
 {% set applicationPath = "/application/" + applicationId %}
 {% set formaction = applicationPath %}
 {% set choices = applicationValue("choices") %}
-{% set title = "Course choices" %}
+{% set title = "Course choices review" %}
```
fofr: How about `Your course choices`?

fofr:
Now we've merged #245, do we still need this?

PR #246, opened 2019-10-22
Content-withdraw part 1

laurahermionetennant92:
Smol changes to the submission page, before working on the instructions on how to amend/withdraw

### DIFF
--- a/app/views/application/confirmation.njk
+++ b/app/views/application/confirmation.njk
@@ -16,12 +16,12 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h2 class="govuk-heading-m">What happens next</h2>
-      <p class="govuk-body">You will receive an email from the Apply for postgraduate teacher training service to confirm the successful submission of your application.</p>
+      <p class="govuk-body">You will get an email from the Apply for teacher training service to confirm the successful submission of your application.</p>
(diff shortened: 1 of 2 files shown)


app/views/application/confirmation.njk
```diff
@@ -16,12 +16,12 @@
...
+      <p class="govuk-body">You will get an email from the Apply for teacher training service to confirm the successful submission of your application.</p>
       <h2 class="govuk-heading-m">References</h2>
-      <p class="govuk-body"> We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 10 working days to respond to your application.</p>
+      <p class="govuk-body"> We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 40 working days to respond to your application.</p>
       <h2 class="govuk-heading-m">Interview</h2>
       <p class="govuk-body">If your training provider decides to offer you an interview, they will get in touch directly via email to organise a date and give you all the information you need.</p>
-      <p class="govuk-body"><a href="/applications">Track the progress of your application</a>.</p>
+      <p class="govuk-body"><a href="/applications">Amending and tracking your application</a>.</p>
```
paulrobertlloyd: @laurahermionetennant92 Should we call out amending here? The previous link feels more correct (with _track_, versus _tracking_; _tracking_ sounds like it might be guidance about how that works).

PR #248, opened 2019-10-22
Remove start page from find flow

fofr:
As per https://docs.google.com/drawings/d/1d30V3qtVYQNL_gWIMuGjgjm2doWMIDyIyYWchIKnPe8/edit

- Remove start page from flow
- Tweak title of found course page
- Use Find course styles to indicate which course was chosen
- Link to course to allow double checking (we saw a user that was uncertain about which course they’d clicked – https://lookback.io/watch/pJKvbAXRyrbweqCHP?t=30m41.91s)

Screen Shot 2019-10-22 at 16 29 45

### DIFF
--- a/app/routes/apply.js
+++ b/app/routes/apply.js
@@ -38,7 +38,7 @@ module.exports = router => {
         courseCode
       }
 
-      res.redirect('/') // Go to Apply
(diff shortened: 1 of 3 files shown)


fofr:
> go straight into sign up flow from Find
Need to think how this will work for existing users.

PR #251, opened 2019-10-23
Proper location choices

fofr:


fofr:
Github isn't letting me update the PR description, which should be:

- Switch from hardcoded list of locations to correct locations for a course
- Handle case when a course only has one location – skip the selection screen and hide the change link
- Correctly propagate the correct location address

PR #274, opened 2019-11-04
Restore course codes for non-opted in course Apply page

paulrobertlloyd:


fofr:
We need to rethink the use of codes on Find and Apply
You don’t really need the provider code for UCAS (lookup is by name)
And the course code (programme) is only useful if there’s multiple courses for the same subject with the same name.

paulrobertlloyd:
Update PR to remove extraneous `courseCodeHtml` variable, leaving design untouched.

PR #307, opened 2019-11-22
Ask if want to add other qualifications

paulrobertlloyd:
Currently, if you don’t want to add any other qualifications, there’s no means of saying so. This means when it comes to reviewing your answers and validating that you have completed all sections, it’s impossible to complete your application without adding a qualification. 🤪

This PR updates the design of the other qualifications section so that we ask a candidate if they want to add any other qualifications. If they select ‘No’, they are returned to the application overview and the section is marked as completed. If they click back into this section, they can select ‘Yes’ and proceed as before.

Screenshot 2019-11-22 at 12 33 05

### DIFF
--- a/app/views/application/other-qualifications/index.njk
+++ b/app/views/application/other-qualifications/index.njk
@@ -1,7 +1,7 @@
 {% extends "_form.njk" %}
-
 {% set applicationPath = "/application/" + applicationId %}
 {% set title = "Other relevant qualifications" %}
+{% set formaction = applicationPath + "/other-qualifications/answer" %}
 
 {% block pageNavigation %}
   {% if referrer %}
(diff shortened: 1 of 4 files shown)


fofr:
Given that most candidates won't need this, we haven't seen anyone who would, it seems heavy handed to introduce a question like this for everyone. 🤔 

I fear that it also sets the wrong precedent by suggesting that many candidates won't be adding their qualifications. Would this disadvantage someone who has that interpretation?

paulrobertlloyd:
Maybe its a case of something like ‘Yes (recommended)’ or similar. But having an optional section be pseudo-required seems very odd.

app/views/application/other-qualifications/index.njk
```diff
@@ -17,103 +17,28 @@
...
-    {{ appAutocomplete({
-      label: {
-        text: "In which country is your institution based?"
-      },
-      items: countries
-    } | decorateApplicationAttributes(["other-qualifications", id, "country"])) }}
-  {% endset %}
+  <p class="govuk-body">Providers use information about your other qualifications to assess your academic ability and the broader contribution you can make to a school.</p>
+  <p class="govuk-body">You can include your GCSEs, A levels and vocational, practical and creative qualifications.</p>
+  <p class="govuk-body">Undergraduate and postgraduate degrees should be included in the ‘Degree’ section.</p>
```
fofr: Can we drop this line?

PR #313, opened 2019-12-02
Capture reference for teacher training candidate

paulrobertlloyd:
### Start page
We ask referees if they are happy to give a reference. If they select no, a free text area allows them to provide a reason.

**Note:** the email we send referees already mentions the respective privacy policy, but we should we should link to it from the email.

01-start

### Decline to give a reference
Should the referee have selected this option in error, we provide contact details to rectify. If this happens a lot, we can think about adding a self-serve way to recover.

01-decline

### Confirm details
Give the referee the opportunity to correct any details. If they select no, a free text area is shown where they can tell us what needs correcting.

02-confirm-details

### Comments
The actual reference content. We provide guidance here about the sort of content that would be helpful here.

03-comments

### Confirmation 
Confirm that their reference has been submitted. At this point, we then ask if they would like to opt-in to user research.

04-confirmation

### Finish
If they opt in, we include a message to say researchers will be in touch (this may need revising based on how we choose to engage with referees)

05-finish-user-research-opt-in

If they decide not to opt-in, we don’t include that message.

05-finish


app/views/reference/decline.njk
```diff
@@ -0,0 +1,10 @@
+{% extends "_content.njk" %}
+
+{% set candidate_name = "Jane Doe" %}
+{% set title = "You’ve declined to give a reference" %}
+{% set hasAccountLinks = false %}
+
+{% block primary %}
+  <h2 class="govuk-heading-m">We will tell {{ candidate_name }} and ask them for a new referee</h2>
+  <p class="govuk-body">If you declined to give a reference in error, please email us at <a href="mailto:becomingateacher@digital.education.gov.uk">becomingateacher@digital.education.gov.uk</a>.</p>
```
fofr: > If you declined in error, email us at …

PR #318, opened 2020-01-08
Refactor reviewing and submitting incomplete application

paulrobertlloyd:
* Fixes review page not being viewable
* Fixes review page not being submittable
* Fixes issue where trying to complete an incomplete GCSE or Degree from the review page, wouldn’t allow you to answer the follow on questions about year, grade etc (`referrer` took precedence. Now only directs to referrer if grade and year already have values).
* Updates CSS and markup for banner component to match that used on line service (including ability to display an icon)
* Uses banner component to show blue/red incomplete warnings on review page (better reflecting component used on live service)
* Updates section review includes to display new incomplete banners
* Updates incomplete messages to match those used on live service
* Updates routes for review pages to validate every section (and moves this logic into a separate file)
* When creating a new `application` object, populates it with empty objects and `null` values

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -141,29 +141,36 @@ module.exports = router => {
 
   // Render review page
   router.get('/application/:applicationId/review', (req, res) => {
-    var app = utils.applicationData(req);
-    var pageObject = {};
-    var successFlash = req.flash('success');
+    var applicationData = utils.applicationData(req)
+    var pageObject = {}
+    var successFlash = req.flash('success')
 
-    if(successFlash[0] == 'submitted-incompleted-application') {
-      pageObject.errorList = [];
+    if (successFlash[0] === 'submitted-incompleted-application') {
+      pageObject.errorList = []
 
-      if(Object.keys(app.choices).length == 0) {
+      if (typeof applicationData.choices === 'undefined' || Object.keys(applicationData.choices).length === 0) {
         pageObject.errorList.push({
-          text: "Courses: section incomplete",
-          href: "#choices-error"
-        });
+          text: 'Course choices not marked as completed',
+          href: '#missing-course-choices'
+        })
       }
 
-      if(Object.keys(app['other-qualifications']).length == 0) {
+      if (typeof applicationData['school-experience'] === 'undefined' || Object.keys(applicationData['school-experience']).length === 0) {
         pageObject.errorList.push({
-          text: "Other relevant qualifications: section incomplete",
-          href: "#other-qualifications-error"
-        });
+          text: 'Volunteering with children and young people is not marked as completed',
+          href: '#missing-school-experience'
+        })
+      }
+
+      if (typeof applicationData['work-history'] === 'undefined' || Object.keys(applicationData['work-history']).length === 0) {
+        pageObject.errorList.push({
+          text: 'Work history is not marked as completed',
+          href: '#missing-work-history'
+        })
       }
     }
 
-    res.render('application/review', pageObject);
+    res.render('application/review', pageObject)
   })
 
   router.post('/application/:applicationId/review', (req, res) => {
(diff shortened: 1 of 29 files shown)


app/routes/application.js
```diff
@@ -10,7 +10,25 @@ function createNewApplication (req) {
 
   if (typeof data.applications[code] === 'undefined') {
     data.applications[code] = {
-      status: 'started'
+      status: 'started',
```
adamsilver: It seems to me that you've added these defaults so that later, in the routes, you can check to see if the user added choices via the following code:

`if (Object.keys(choices).length === 0)`

I don't think we should define empty objects here. Instead I'd make the check:

`if(application.choices)`

If the fact it's empty is a little more complicated (which I think it might be judging from the fact that some properties are set to an empty object and some are set to null) then I'd suggest a method to encapsulate this so I can do:

`if(app.isTaskComplete('choices'))`

Behind the scenes ,this function might hand off to `app.isChoicesComplete()`. 

`

app/routes/application/review.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+module.exports = router => {
+  router.get('/application/:applicationId/review', (req, res) => {
+    var applicationData = utils.applicationData(req)
+    var pageObject = {}
+    var successFlash = req.flash('success')
+
+    if (successFlash[0] === 'submitted-incompleted-application') {
+      pageObject.errorList = []
+      const sections = {
```
adamsilver: This now looks like if the user submits an incomplete application then every single thing is marked as incomplete instead of just the incomplete things.

app/routes/application/review.js
```diff
@@ -0,0 +1,118 @@
...
+    const reasonableAdjustments = applicationData['reasonable-adjustments']
+    const schoolExperience = applicationData['school-experience']
+    const subjectKnowledge = applicationData['subject-knowledge']
+    const workHistory = applicationData['work-history']
+
+    // User tried to submit incomplete application
+    // just checking choices are empty for now
+    // to make real need to check over every section
+    if (
+      Object.keys(choices).length === 0 ||
```
adamsilver: Instead of all these checks, I think this should be a method on the application like this:

`if(!app.isComplete())`

app/routes/application/review.js
```diff
@@ -0,0 +1,118 @@
...
+      Object.keys(science).length === 0 ||
+      Object.keys(personalStatement).length === 0 ||
+      Object.keys(subjectKnowledge).length === 0 ||
+      Object.keys(interview).length === 0 ||
+      Object.keys(referees).length === 0
+    ) {
+      req.flash('success', 'submitted-incompleted-application')
+      res.redirect(`/application/${req.params.applicationId}/review`)
+    } else {
+      console.log('applicationData.status', applicationData.status)
```
adamsilver: Rogue log

app/views/_components/banner/template.njk
```diff
@@ -1,14 +1,4 @@
 <div class="app-banner{% if params.type == "success" %} app-banner--success{% elif params.type == "warning" %} app-banner--warning{% endif %}{{- " " + params.classes if params.classes}}" {%- for attribute, value in params.attributes -%} {{ attribute }}="{{ value }}"{% endfor %}>
-{% if params.icon != false %}
```
adamsilver: I think the icons are valuable and that we should show them in order to provide a comparable experience to users who don't see colour so well.

paulrobertlloyd: Have restored the ability to include icons with the banner component, but disabled to main parity with what’s shipped (not adding/changing features in this PR, just making things work!)

app/routes/application/review.js
paulrobertlloyd: Created a new utility function `hasCompletedApplication(req)`, which follows a similar design to the other checks we have.

app/routes/application/review.js
paulrobertlloyd: With the changes made so far, behaviour appears correct; the errors shown at the top of the page reflect the incomplete sections. So may skip refactoring this.

app/routes/application.js
paulrobertlloyd: Created a new utility function `hasCompletedSection(key)`, which checks for either `null` or empty keys.

paulrobertlloyd:
@adamsilver If you could take another look over this, that’d be grand. Also fixed an issue where trying to complete an incomplete GCSE or Degree from the review page, wouldn’t allow you to answer the follow on questions about year, grade etc. Think I’ve fixed that!

fofr:
Nice work! A much needed update to the prototype.

paulrobertlloyd:
@adamsilver @fofr Thanks both — this thing is a monster! 🤪

PR #321, opened 2020-01-13
Include field for capturing a part time working pattern

fofr:
There’s a fair difference between a 1hr per week Saturday job and a 4 day per week part time one.
Providers have said they need to know the difference when looking at a candidate’s work history.

Also:
- renames "Type" to "Working pattern" in review
- removes hyphenations

Screen Shot 2020-01-13 at 11 47 16

Screen Shot 2020-01-13 at 11 49 52

https://trello.com/c/4B34Ikal/745-re-consider-how-we-frame-part-time-and-full-time-work-following-feedback-from-providers


paulrobertlloyd:
:+1: ‘working pattern… works really well in review.

PR #328, opened 2020-01-29
Add design for a referee feedback form after giving a reference

fofr:
localhost_3000_referee-confirmation (2)
localhost_3000_referee-confirmation (1)

https://trello.com/c/aaKMc14Y/800-design-questionnaire-to-get-feedback-from-referees

### DIFF
--- a/app/views/referee-confirmation.njk
+++ b/app/views/referee-confirmation.njk
@@ -8,7 +8,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h2 class="govuk-heading-l">Tell us about your experience giving a&nbsp;reference</h2>
+      <h2 class="govuk-heading-l">Tell us about your experience of giving a&nbsp;reference</h2>
       <p class="govuk-body">Your feedback will help us to improve our service.</p>
 
       {% set whyRatingHtml %}
@@ -25,7 +25,7 @@
             text: "Tell us why you chose that rating"
           },
           hint: {
-            text: "Did you have questions that weren’t answered by the guidance?"
+            text: "Did you have questions that were not answered by the guidance?"
           }
         } | decorateApplicationAttributes(["referee-confirmation"])) }}
       {% endset %}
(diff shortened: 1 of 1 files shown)


fofr:
For reviewers: Markup has problems with the repeated conditional HTML fields, but as this is an illustrative design that doesn't fit into a user journey within the prototype, I figured this would be fine.

adamsilver:
Nice one @fofr. In addition to what I said on Slack I'd suggest:

- not using the word ‘please’
- using text labels instead of numbers (and not a mix)
- looking if we could combine the heading and paragraph to something like ‘Tell us about your experience so we can improve the way giving a reference works’
- considering marking (some of) the questions as optional

fofr:
> not using the word ‘please’
Discussed this with @emmajhf  and Kahar, and we think this is a good exception to https://www.gov.uk/service-manual/design/writing-for-user-interfaces#tone because this is an optional survey and only beneficial to us.

> using text labels instead of numbers (and not a mix)
Good idea. @emmajhf what words would you suggest? (Very poor, Poor, Ok, Good, Very good?)

app/views/referee-confirmation.njk
```diff
@@ -0,0 +1,132 @@
...
+
+{% block content %}
+  {{ govukPanel({
+    classes: "govuk-!-margin-bottom-8",
+    titleHtml: "Your reference for Jane Doe has been&nbsp;submitted"
+  }) }}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h2 class="govuk-heading-l">Tell us about your experience giving a&nbsp;reference</h2>
```
fofr:
```suggestion
      <h2 class="govuk-heading-l">Tell us about your experience of giving a&nbsp;reference</h2>
```

app/views/referee-confirmation.njk
```diff
@@ -0,0 +1,132 @@
...
+        } | decorateApplicationAttributes(["referee-confirmation"])) }}
+      {% endset %}
+
+      {% set whyGuidanceRatingHtml %}
+        {{ govukTextarea({
+          label: {
+            text: "Tell us why you chose that rating"
+          },
+          hint: {
+            text: "Did you have questions that weren’t answered by the guidance?"
```
fofr:
```suggestion
            text: "Did you have questions that were not answered by the guidance?"
```

emmajhf:
@fofr, I think those labels (Very poor, Poor, Ok, Good, Very good) work well. I agree about not needing numbers.

PR #332, opened 2020-02-05
Adapt adding new referee pattern to adding 2 referees

fofr:
See documentation in https://github.com/DFE-Digital/bat-design-history/pull/53

01-you-need-to-add-2-new-referees
02-add-your-first-new-referee
03-add-your-second-new-referee
04-check-each-referees-details
05-application-dashboard-after-adding-referees
06-application-dashboard-with-warning

### DIFF
--- a/app/views/application/references/alternative-first.njk
+++ b/app/views/application/references/alternative-first.njk
@@ -1,0 +1,64 @@
+{% extends "_form.njk" %}
+
+{% set type = applicationValue(["referees", id, "type"]) | lower %}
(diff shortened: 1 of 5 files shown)


adamsilver:
This is great.

My only suggestion is to use the red alert at the top above the h1 rather than the exclamation mark text as you have it. I think it's way hard to spot.

fofr:
@adamsilver Agree, I've updated the code and the screenshots in the design history

adamsilver:
Merge!

PR #333, opened 2020-02-05
Updates to eligibility questions

fofr:
- Remove confusing note about EEA (which means Lichtenstein, Norway or Iceland)
- Clarify why we are asking these questions – that it's us that's not ready, rather than them

Screen Shot 2020-02-05 at 10 36 53

https://trello.com/c/EOZDqPD7/888-preventing-candidates-getting-blocked-by-the-eligibility-questions

### DIFF
--- a/app/views/account/eligibility.njk
+++ b/app/views/account/eligibility.njk
@@ -11,7 +11,7 @@
 {% endblock %}
 
 {% block primary %}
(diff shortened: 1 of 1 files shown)


emmajhf:
Hi @fofr 

This looks great. I have a question about 'if you do not meet the criteria below'. It still sounds like candidates might fail somehow. I might be overthinking this, but I wonder if we can stress the 'it's not you – it's us' thing even more. How about:

Check we're ready for you to use this service

This service is new and not available to everyone yet. Check if we're ready for you by answering the questions below.

If you can't use our service, you can still apply for teacher training through UCAS.

PR #338, opened 2020-02-19
Non-functioning full/part time design

fofr:
@paulrobertlloyd Is it worth wiring this up properly, or is this branch illustration enough?

Screen Shot 2020-02-18 at 20 59 43

Screen Shot 2020-02-18 at 21 04 43


paulrobertlloyd:
@fofr Perhaps we should we include it in a broader entry about how we deal with courses that are full?

PR #339, opened 2020-02-20
Update content on UCAS apply pages

paulrobertlloyd:
Updates the design of the page that candidates see if they select a course that’s not on Apply (within the Apply ‘choose a course’ journey). We may need to tweak the content to reassure candidates that they can use both services (picking three courses across both) if they so wish. /cc @emmajhf 

Apply on UCAS


emmajhf:
Hello @paulrobertlloyd, thanks for this. I've made some suggestions for the content in https://docs.google.com/document/d/1J5n_EJQ1rVaAUo8lcZ1sa-ucrr4a5F6_-SwHah-JYu0/edit?usp=sharing

fofr:
Can "Return to your application" be a link?

PR #343, opened 2020-02-20
Alternative content for before you continue page

paulrobertlloyd:
@emmajhf Me and @fofr played around with this page to make it less wordy and put the emphasis on the one message we want users to take away from this page, that we recommend that they choose a course first. Thoughts?

Before you continue


emmajhf:
@paulrobertlloyd I think this looks good. Just two suggested changes to contractions:

1) In the second sentence, change 'it is available' to 'it's available'.
2) In the third sentence, change 'which isn't' to 'which is not'.

(GDS style is to avoid negative contractions.)

PR #344, opened 2020-02-20
declaring any safeguarding issues

EmmaFrith:
changed content following candidate user research

### DIFF
--- a/app/views/application/suitability/index.njk
+++ b/app/views/application/suitability/index.njk
@@ -1,6 +1,6 @@
 {% extends "_form.njk" %}
 
-{% set title = "Your suitability to work with children" %}
+{% set title = "Declaring any safeguarding issues" %}
 
 {% block pageNavigation %}
   {% if referrer %}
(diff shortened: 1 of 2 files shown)


app/views/application/suitability/index.njk
```diff
@@ -45,17 +48,17 @@
...
+      text: "Do you want to share any safeguarding issues?",
       classes: "govuk-fieldset__legend--m"
     }
   },
   items: [{
     value: "true",
     text: "Yes, I want to share something",
     hint: {
-      text: "You can tell us about it here or get in touch with your training provider directly."
+      text: "You talk to your provider directly or tell us about it here. This won’t stop you from submitting your application."
```
fofr: Do we need the "This won’t stop you from submitting your application" bit?

EmmaFrith: This was because there were concerns from candidates about whether they'd still be able to continue with their application

PR #346, opened 2020-02-21
Revised pilot eligibility flow

paulrobertlloyd:
Basis for revised ‘check you can use this service’ flow, that assesses both _eligibility_ and _suitability_ giving recommendations based on the answers given.

Placeholder content by @paulrobertlloyd is dire; @emmajhf my apologies.


fofr:
Looking good :100:

PR #349, opened 2020-02-21
Revised Find interstitial

paulrobertlloyd:
Just the plumbing to make the Find interstitial work with three options. Needs *a lot* of refinement and review… even deciding if we want to use this approach.

choose-apply-service

### DIFF
--- a/app/routes/apply.js
+++ b/app/routes/apply.js
@@ -1,6 +1,6 @@
 module.exports = router => {
   // const dualRunningDefaultVarient = Math.floor(Math.random() * 3) + 1;
-  const dualRunningDefaultVarient = 1
+  const dualRunningDefaultVarient = 4
 
   router.get('/apply/:providerCode/:courseCode', (req, res) => {
     const variant = req.query.variant || dualRunningDefaultVarient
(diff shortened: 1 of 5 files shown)


app/views/apply/_variant-choice-with-sign-in.njk
```diff
@@ -0,0 +1,33 @@
+<p class="govuk-body">{{ serviceName }} is a new GOV.UK service that will eventually replace UCAS, but it is not suitable for everyone yet.</p>
```
benilovj: > Apply for teacher training is a new GOV.UK service...

Just my 2p: this may read strangely if you don’t know that ‘Apply for teacher training’ is a service name, as opposed to an imperative to apply.

paulrobertlloyd: Yup, there’s a fine line between you can I use _Apply for teacher training_ (the service), and can I _apply for teacher training_. This is where naming services to describe the activity falls apart, even more so when there are two services that do the same thing.

We’ll be looking at this more closely next week when @emmajhf is back in.

emmajhf:
Hi @paulrobertlloyd, what do you think of the following:

**Do you want to apply for this course using a new GOV.UK service?**

The course you’ve chosen is available through a new GOV.UK service called Apply for teacher training. This is designed to make it easier to apply for teacher training.

For now, the new service is only available for some courses. 

If you have other training providers who aren’t yet signed up for the service, you may also need to apply through UCAS.

Check if the new service is right for you
Sign back in to Apply for teacher training
Apply through UCAS


I'm not sure the buttons option is working yet, so I've also suggested another option using no buttons:
https://docs.google.com/document/d/1N9fZYjd7B_ZNrErinvYEkh-Z6OQrC65tHqMZzk-JvRw/edit?usp=sharing

PR #352, opened 2020-02-26
Reinstate reference types with revised content

paulrobertlloyd:
choosing-your-referees
first-referee
details
review


emmajhf:
Looks good. One tiny change: can we replace 'supply a reference' with 'give a reference'.

PR #355, opened 2020-02-27
Updated reference flow

paulrobertlloyd:
start

relationship

safeguarding

review

* Probably want to think a bit more about the start page content.
* If you select ‘Yes’ for ‘Do you know of any reason…’ question, we don’t ask for a reference, but ask for a reason inline. Is that the right approach? Alternatively, could still ask for comments, but change the question slightly.
* Still need to add adapting guidance on comment page based on type of reference being asked for.


adamsilver:
This looks really good @paulrobertlloyd. Some comments/suggestions:

1. Can you make the textarea smaller in height?

2. For the relationship confirmation page (page 2) how about this:

```
# Confirm your relationship to Jane Doe

Jane Doe describes your relationship as follows:

> [Description]

Legend: Is this an accurate description of your relationship?
( ) Yes
( ) No
[Continue]

3.  Why does the referee need to know who the candidate applied for on the start page (or any page)?

4. Page 3 is framed in the negative so ‘Yes’ is actually a bad thing. Is this on purpose?

5. I think the check answers page needs to reflect all the answers as the referee supplied them in. For example:

```
Is this an accurate description of your relationship?:  No  Change
Actual relationship: [Blah blah blah]  Change
...
```

fofr:
This looks great.

One reason a referee might need to know who they applied to is because one of those providers might contact them about the reference, for a follow up.

Agree with Adam about having “Yes” in the safeguarding bit, content tweak might say “I have a reason” or similar.

I like the relationship page, I think Adams suggestions make sense.

I like the idea of short cutting the reference if they have a reason they aren’t safe to work with children. Risky though. But I guess it depends on the reason, a provider might dismiss the reason and then be missing a reference, or a full reference might corroborate or give a fuller view.

emmajhf:
Hi @paulrobertlloyd, thanks for this. Looks great.

Comments from me:

- I'd change 'Give a teacher training reference for Jane Doe' to 'Give a reference for Jane Doe'. The referee has already had the email and been taken to a site called 'Apply for teacher training'.

-  I think it's worth including details of what the candidate applied for. Whenever I get asked to give a reference, I'm interested in the role the person will be taking on.

- On the first page, can we change 'Start now' to 'Give a reference'?

- For the title of the relationship page, we could use part of Adam's suggested text. I like: 'Confirm how you know Jane Doe'. Introductory sentence could be: 'Jane Does describes your relationship as follows:'

- For the 'Do you know of any reason' page, can we have 'No' first? Presumably, most people will choose this. With the hint text for 'yes', how about: 'Tell us why you think Jane Doe should not work with children'. I agree there's no point in asking for a full reference if they think the candidate shouldn't work with kids.

PR #359, opened 2020-03-05
Structured other qualifications

paulrobertlloyd:
First pass at prompting for a qualification type (GCSE, A Level, AS Level, Other) within the other qualifications section.

Also includes support for entering structured subjects for GCSE qualifications. This aspect may need more work (ie do we show pilot qualifications, do we merge qualifications with similar subject titles, etc). These issues can be addressed later, in a separate PR.

other-qualification-types

### DIFF
--- a/app/views/application/other-qualifications/index.njk
+++ b/app/views/application/other-qualifications/index.njk
@@ -1,7 +1,7 @@
 {% extends "_form.njk" %}
 
 {% set applicationPath = "/application/" + applicationId %}
-{% set title = "Other relevant qualifications" %}
+{% set title = "Academic and other relevant qualifications" %}
 
 {% block pageNavigation %}
   {% if referrer %}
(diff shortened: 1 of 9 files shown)


app/views/application/other-qualifications/index.njk
```diff
@@ -17,75 +17,73 @@
...
+        text: "Enter type of qualification"
+      }
+    } | decorateApplicationAttributes(["other-qualifications", id, "type-uk"])) }}
+  {% endset %}
 
-  {% set internationalConditionalHtml %}
+  {% set nonUkConditionalHtml %}
+    {{ govukInput({
+      label: {
+        text: "Enter type of qualification"
```
fofr:
```suggestion
        text: "Type of qualification"
```
Do we need "Enter"?

app/views/application/other-qualifications/index.njk
```diff
@@ -17,75 +17,73 @@
...
+      value: "as-level",
+      text: "AS level"
+    }, {
+      value: "a-level",
+      text: "A level"
     }, {
-      value: "international",
-      text: "Outside the UK",
+      value: "other-uk",
+      text: "Other UK qualification",
```
fofr:
```suggestion
      text: "Other",
```
Do we need to avoid international here?

PR #360, opened 2020-03-11
Additional guidance for GCSE grades

paulrobertlloyd:
## Minimum qualification grade

For GCSE or equivalent grades, include guidance around minimum grade and suggestion to contact a provider if currently don‘t meet the criteria.

Screenshot 2020-03-11 at 11 39 05

* * *

## Multiple English qualifications

For English GCSE subject, add two disclosures:
1. Only have an English Literature qualification
2. Have more than one English qualification 

Screenshot 2020-03-11 at 11 38 50

### DIFF
--- a/app/views/_includes/item/gcse.njk
+++ b/app/views/_includes/item/gcse.njk
@@ -23,7 +23,7 @@
     },
     actions: {
       items: [{
-        href: applicationPath + "/gcse/" + item.id + "?referrer=" + referrer,
+        href: applicationPath + "/gcse/" + id + "?referrer=" + referrer,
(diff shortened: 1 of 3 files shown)


fofr:
"Provider expect you to", shouldn't it be "To train you must"?

emmajhf:
@paulrobertlloyd @fofr I prefer 'Providers expect you to' over 'To train you must' because the latter's not strictly true, and I worry it would be off-putting. We know there are cases where people can still train if they have a D (or whatever) because they have other evidence of their skills.

I think this is good to merge now.

PR #361, opened 2020-03-13
Allow creation of Apply 2 applications

fofr:
- Include an "apply2" boolean for each application, which we can use to toggle Apply2 specific features (eg limit that application to 1 course choice)
- Create an apply2 dummy application derived from the existing dummy
- Add a route for copying an existing application and making it into a new Apply 2 one
- Add missing suitability fields to defaults and dummy applications

### DIFF
--- a/app/data/dummy-apply-2-application.js
+++ b/app/data/dummy-apply-2-application.js
@@ -1,0 +1,14 @@
+// Copies dummy application and tweaks some fields
+var dummyApply2Application = JSON.parse(JSON.stringify(require('./dummy-application')))
+
+// Custom Apply 2 fields
+dummyApply2Application.apply2 = true
+dummyApply2Application.previousApplications = ['12345']
+
+// No sections completed
+dummyApply2Application.completed = {}
+
+// No choices yet
+dummyApply2Application.choices = {}
+
+module.exports = dummyApply2Application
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
Should the `apply2` boolean be a feature flag?

app/data/dummy-apply-2-application.js
```diff
@@ -0,0 +1,14 @@
+// Copies dummy application and tweaks some fields
+var dummyApply2Application = JSON.parse(JSON.stringify(require('./dummy-application')))
+
+// Custom Apply 2 fields
+dummyApply2Application.apply2 = true
+dummyApply2Application.previousApplications = ['12345']
+
+// No sections completed
+dummyApply2Application.completed = {}
```
paulrobertlloyd: `completed` or `reviewed`. Same to ensure parity with Apply 1 application?

app/data/dummy-apply-2-application.js
```diff
@@ -0,0 +1,14 @@
+// Copies dummy application and tweaks some fields
+var dummyApply2Application = JSON.parse(JSON.stringify(require('./dummy-application')))
```
paulrobertlloyd: Smart!

app/data/dummy-apply-2-application.js
fofr: I was going to use the same data structure but change the language in the UI, so we reuse logic.

PR #364, opened 2020-03-17
add guidance warning candidates that they'll be contacted by provider…

EmmaFrith:
…s about anything they declare

add guidance warning candidates that they'll be contacted by providers about anything they declare. moved some of the hint text up into body copy because hint text was getting too long

### DIFF
--- a/app/views/application/suitability/index.njk
+++ b/app/views/application/suitability/index.njk
@@ -16,24 +16,24 @@
 {% endblock %}
 
 {% block primary %}
-<p class="govuk-body">Teacher training providers need to make sure that it’s safe for you to work with children and young people.</p>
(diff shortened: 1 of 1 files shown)


app/views/application/suitability/index.njk
```diff
@@ -53,7 +53,7 @@
     value: "true",
     text: "Yes, I want to share something",
     hint: {
-      text: "Talk to your provider directly or tell us about it here. This won’t stop you from submitting your application."
+      text: "Talk to your provider directly or give information here. If you share something here, your provider may get in touch to talk about it."
```
fofr: > "give information here. If you share something here"

Too many "here"s?

PR #365, opened 2020-03-19
Align tag colours and statues with what’s in production

paulrobertlloyd:
Possibly bit of a fools errand, but trying to align statuses and tag colours (using new Design System values) with what we have built.


app/filters.js
```diff
@@ -127,5 +127,42 @@ module.exports = (env) => {
...
+        return 'govuk-tag--blue'
+      case 'Offer withdrawn':
+        return 'govuk-tag--red'
+      case 'Application withdrawn':
+        return 'govuk-tag--orange'
+      case 'Declined':
+        return 'govuk-tag--orange'
+      case 'Rejected':
+        return 'govuk-tag--red'
+      case 'Conditions not met':
```
fofr: This one is currently dark blue on prod 😬

paulrobertlloyd: Oh, I had it down as `Conditions met`. There are definitely some funky colours hanging around. I think we need to take a more holistic look.

PR #368, opened 2020-03-20
Show previous applications on dashboard when in Apply 2

paulrobertlloyd:
Trying to lean on existing components, and introduce the timeline component to the candidate interface.

* Wondering what information we need to show about each choice. On production, we now show even more info per choice, should all that be shown as well? Given you can also see choices when you view an individual application, perhaps not?

* Is this ordering right? Should it be latest application first, or first application first?

### Previous applications on dashboard

One previous application:

Screenshot 2020-03-23 at 16 37 30

Many previous applications:

Screenshot 2020-03-23 at 16 39 06

### Submitted application

Just tweaked the presentation of the submitted date to align with presentation used elsewhere (not using summary list, but showing hint text below title)

* Should we explicitly label applications with an ordinal, i.e. first application, third application etc?

previous-submitted-application


app/assets/sass/components/_timeline.scss
```diff
@@ -0,0 +1,116 @@
...
+
+.app-timeline__date {
+  @include govuk-font($size: 16);
+  margin-top: govuk-spacing(1);
+  margin-bottom: 0;
+}
+
+.app-timeline__description {
+  @include govuk-font($size: 19);
+  margin-top: govuk-spacing(2);
```
paulrobertlloyd: Changed this from 4 to 2. Does that break things on the provider side, @adamsilver?

fofr:
Thoughts:

- Should the timeline show the most recent application first? (Reverse chronological)
- Rather than first/second/third, we could label with date, "Application you submitted on 1 March 2020" or similar
- Do we need the timeline, would a list of links to each "View application" page be enough?

app/assets/sass/components/_timeline.scss
adamsilver: @paulrobertlloyd we don't have descriptions just yet so no, not yet :)

paulrobertlloyd: @adamsilver Never mind, not using the timeline component now!

fofr:
This PR didn't work, reverted in 8025de9 so we can test.

Looks like styles and data weren't committed.

paulrobertlloyd:
Oooh, I think I know what broke it, one sec!

PR #373, opened 2020-04-02
Group list of providers by region

paulrobertlloyd:
* `eastern` `region_code` to use title ‘East of England’
* If no providers in a region, still show it, but say ‘No providers are available in this region yet.’
* Use ‘No region’ as heading for providers without a `region_code`


laurahermionetennant92:
Hi @paulrobertlloyd  – still think 'No region' sounds weird. If we don't have a region because they span a number of regions, please let's use the heading 'Cross-region providers'.
Was Steve going to find out what other reason there might be for a lack of reason code?

paulrobertlloyd:
@laurahermionetennant92 ‘No region’ shouldn’t appear - each provider has a region. Phew!

PR #374, opened 2020-04-06
Apply 2 updates

adamsilver:
@paulrobertlloyd this PR is increasing in size and taken me a fair amount of work so will do rest of the ticket in separate PR.

Application page:
image

Individual tasks now have checkboxes:
image

Review page now reflects those state for each task:
image

### DIFF
--- a/app/data/dummy-application.js
+++ b/app/data/dummy-application.js
@@ -35,11 +35,7 @@ module.exports = {
     }
   },
   completed: {
-    choices: ['true'],
-    degree: ['true'],
-    'other-qualifications': ['true'],
-    'school-experience': ['true'],
-    'work-history': ['true']
+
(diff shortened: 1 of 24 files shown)


paulrobertlloyd:
Functionally, this seems to work great! Labelling wise, not so sure. But feel to create a new PR for that.

paulrobertlloyd:
Actually, ‘Not started’ is not shown when creating a new Apply 1 application (small thing right now, but worth pointing out)

fofr:
Why was Previous applications moved below Need help?

paulrobertlloyd:
> Why was Previous applications moved below Need help?

Because there’s no evidence of candidates wanting to see previous applications, but they are more likely to want help.

PR #377, opened 2020-04-07
Change last task into a button

adamsilver:
image


paulrobertlloyd:
I don’t think we had a heading last time we tried using a submit button, so certainly worth testing. That said, it does seem a bit repetitive.

adamsilver:
It is a little repetitive as is the previous task. I guess it's hard when a section has only one task.

PR #378, opened 2020-04-15
Structured degrees

paulrobertlloyd:
## 1. Type

Shown below is the version that supports international degrees. Without international features, the radio options disappear.

* Structured options are provided for UK degree type (full names are shown in the autocomplete).
* When the first degree is being added, only undergraduate type degrees are shown.

type

## 2. Subject

Free text answer.

subject

## 3. Institution

Structured answers given **only** for UK degrees.

* Question: we should add an ‘other’ option, but not sure what is the best practice for this – the autocomplete doesn’t allow for entries not provided in the data list. Any thoughts @adamsilver? 

org

## 4. Grade

Different options shown for:
* A UK undergraduate degree
* A UK postgraduate degree
* An international degree

There are quite a lot of options here, most of which will unlikely to be selected. Should we hide them, or allow them to be added as free text within ‘Other’?

grade-undergraduate

grade-postgraduate

grade-international

## 5. Dates

Added start date. The second label changes to *When will you graduate?* if candidate is still studying for this degree. Are these the right labels do you think @EmmaFrith? 

year

## Review

UK undergraduate, UK postgraduate and international degrees shown.

* I have added ‘(Hons)’ to the card title if a UK undergraduate degree, and grade has honours.

review


adamsilver:
There seems to be a bit of a big gap between label and input on some screens.

paulrobertlloyd:
@adamsilver That old chestnut 😉 Seems inconsistent to go from gaps to no gap though.

adamsilver:
Lol, oh it was on purpose 😂

EmmaFrith:
Looking great @paulrobertlloyd! Here's some thoughts on the content from me: 

1. On the undergraduate degree, we have 'what class is your degree', whereas in the Masters degree we have 'what grade is your degree'.

Is there a reason for them being different? On the summary/review page, we say 'grade' in both instances.

(I haven't checked every degree type in the list to check consistency - I only looked at BA and MA).

2. For the 'add undergraduate degree' page - I know this content has been there for a while - but could we go for something more informal?

For example: 'You need an undergraduate to apply for teacher training. If you have further degrees, you can add them after.'

I don't think we need 'Enter the details of your degree as they appear...' because we're structuring them now. Or that bit could be hint text under 'Non-UK degree'?

3. On the review page, I think we can just say 'Institution' rather than 'Institution studied at'

4. On the review page, is it just me or does 'Dates studied' sound a bit odd? Maybe it's because you can't study a date. Could just have 'Dates'?

5. On the review/summary page, I notice we've got 'BSc (Hons) Biology' - but honours degrees don't seem to feature in the list of degrees - if a candidate has an honours degree and wants to state this, is this something they'd need to put in the 'subject' page?

6. On the review/summary page, 'country studied in' could just be 'location'

7. The page 'Where did you study your degree?' could be misleading and encourage people to search for location not institution. Not all unis have a place in their name. Maybe it'd be better to go for 'What institution did you go to?'

paulrobertlloyd:
Thanks @EmmaFrith! I have integrated all you suggestions where relevant.

PR #380, opened 2020-04-20
Reason for withdrawal

paulrobertlloyd:
### Withdraw (no change)

withdraw

### Withdraw confirmation

After you withdraw, now show a confirmation screen, which asks 2 questions:

* Do you have a reason for withdrawing?
* Can we contact you about your experience of withdrawing?

We make it clear that feedback will only be shared with the Department for Education, and reinforce this again if candidate selects ‘Yes’ to the first question.  

withdraw-confirmation

Selection ‘Yes’ to either, show the following free text areas:

withdraw-confirmation-options

### DIFF
--- a/app/data/session-data-defaults.js
+++ b/app/data/session-data-defaults.js
@@ -22,7 +22,6 @@ module.exports = {
     nationality: process.env.MVP !== 'true',
     self_amend: process.env.MVP !== 'true',
     self_amend_contact_details: process.env.MVP !== 'true',
-    self_withdraw: process.env.MVP !== 'true',
     interview_preferences: true
   },
   // applications: (process.env.PRODUCTION === 'true') ? {} : testData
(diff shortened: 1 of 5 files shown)


EmmaFrith:
Looking good @Paul Lloyd

Comments/ questions:

- Why do we add the extra clause (in brackets) here?: 'If you withdraw this course choice (and all your other course choices), you....' - this is confusing because the candidate is about to withdraw 1 course, not all of them. 

- On the reason for withdrawing page: why are we not giving check boxes with possible withdrawal reasons? It'd be much more useful for us to have structured data - and I think would make it more likely that the candidate would complete it. 

- 'Can we contact you about your experience of withdrawing?' - makes it sound like a huge ordeal. What exactly do we want to find out from them when we contact them and why?

paulrobertlloyd:
> …why are we not giving check boxes with possible withdrawal reasons?

For now we want to capture free text. As we learn more about reasons, we can start to capture structured answers, but for now we don’t want to provide leading answers.

PR #383, opened 2020-04-22
Improvements to adding/amending referees when applying the first time and applying again

adamsilver:
Changes:
- Clicking the task from the task list takes the user straight to add their first or second referee (no interstitial)
- Button states (shown below)
- Removed the line that says ‘we'll contact your referees after you submit your application...’. There’s a note in context within the summary box (shown below)

# Button state when both referees have been added

image

# Button state when no referees have been added

Note: this page is only seen when you delete both referees. When you click the task on from the task list page, you’ll be taken to the first referee form to add the referee.

image

# Button state when one referee added

image

# Not requested state

image

# Confirmation page for applying again

- the references section won’t appear if the user never changed a referee
- the content changes if the user changed 1 referee vs both referees

image


app/views/application/references/review.njk
```diff
@@ -39,6 +42,17 @@
       text: "Continue without adding second referee",
       href: applicationPath
     }) }}
+  {% elseif not applicationValue(["referees", "first"]) %}
+    {{ govukButton({
+      text: "Add a first referee",
```
fofr: "Add a first referee" feels a little clunky.

fofr: When does this situation arise?

fofr:
When you delete your first referee you see this:

Screen Shot 2020-04-22 at 11 08 32

I don't think this is right. I think the "Add a first referee" content is unnecessary.

When you delete your first referee, the existing one should become the first, and you should be prompted to "Add another" rather than "Adding a first" or "Adding a second". Whether a referee is first or second is inconsequential.

app/views/application/confirmation.njk
```diff
@@ -18,9 +18,19 @@
...
+        <h2 class="govuk-heading-m">References</h2>
+
+        {% if applicationValue(["referees", "first", "status"]) == "Received" or applicationValue(["referees", "second", "status"]) == "Received" %}
+          <p class="govuk-body">We’ll contact your referee to ask for your reference. When we receive it, we’ll add it to your application and send it to your training provider(s).</p>
+          <p class="govuk-body">We can’t send your application to your training provider(s) without your reference.</p>
+        {% else %}
+          <p class="govuk-body">We’ll contact your referees to ask for your references. When we receive them, we’ll add them to your application and send it to your training provider(s).</p>
+          <p class="govuk-body">We can’t send your application to your training provider(s) without your references.</p>
+        {% endif %}
+      {% endif %}
```
fofr, line 32: Nice. I like that this whole section gets removed when not necessary.

adamsilver:
@fofr I’ve fixed the button labels. Much better thanks.

> When you delete your first referee, the existing one should become the first, and you should be prompted to "Add another" rather than "Adding a first" or "Adding a second". 

Agreed—this is an existing bug in the prototype that I’ll leave for a separate PR.

paulrobertlloyd:
This is all looks great – just curious how we ensure we capture all these nuances for development; a separate design history entry? I take it these changes apply to both Apply 1 & 2?

adamsilver:
Thanks Paul-Yeah, a new history entry and some dev tickets.

PR #384, opened 2020-04-22
Structured degrees iteration

paulrobertlloyd:
* Show undergraduate degree grades for postgraduate degree types
* Allow Level 7 degree types to be selected for first degree (i.e. bachelors and masters, but not doctorates)
* Allow other degree types to be entered
* Improve autosuggest UX

### DIFF
--- a/app/data/degree.js
+++ b/app/data/degree.js
@@ -1642,7 +1642,7 @@ module.exports = function () {
   return {
     types: {
       all: types,
-      undergraduate: types.filter(type => type.level === 6)
+      undergraduate: types.filter(type => type.level === 6 || type.level === 7)
     },
     subjects,
     orgs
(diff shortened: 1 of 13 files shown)


fofr:
I wonder if we need to do better with fuzzy matching?

Screen Shot 2020-04-23 at 08 57 55

fofr:
In other places, "Other" gives a free text field, which is what I expected.

Screen Shot 2020-04-23 at 09 00 50

EmmaFrith:
> I wonder if we need to do better with fuzzy matching?
> 
> (I think I prefer "Which institution did you study at?", rather than "go to", to cover online degrees)
> 
> Screen Shot 2020-04-23 at 08 57 55

Good point @fofr - I'm happy with that

EmmaFrith:
> In other places, "Other" gives a free text field, which is what I expected.
> 
> Screen Shot 2020-04-23 at 09 00 50

I was also expecting other to be free text

PR #386, opened 2020-04-29
Tweaks for accepting or declining offers

fofr:
“Are you sure you want to accept offer” contained guidance on withdrawing other choices that’s not relevant when applying again (there’s only 1 choice).

“Are you sure you want to decline offer” contains guidance about applying again through UCAS, when this feature is launched apply again will not be specific to UCAS.

The updated content removes mention of:

- UCAS when suggesting a candidate can apply again after declining
- the caveat about “all other offers” when declining an offer on an application with 1 choice
- automatically withdrawing other choices when accepting an offer on an application with 1 choice

These changes should also apply to first applications that have only 1 choice.

### DIFF
--- a/app/views/application/decision/accept.njk
+++ b/app/views/application/decision/accept.njk
@@ -12,11 +12,15 @@
 {% block primary %}
   {% include "_includes/item/offer.njk" %}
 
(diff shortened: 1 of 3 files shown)


fofr:
Guidance also added to withdraw screens.

PR #389, opened 2020-05-22
Structured answers for other qualifications

paulrobertlloyd:
## Suggested other UK qualifications and UK vocational grades

If you select ‘Other UK qualification’ suggestions are given for common types of qualification:
* BTEC
* GCE O Level
* NVQ
* Scottish National 5
* Scottish Higher
* Scottish Advanced Higher
* SVQ
* VRQ

Suggesting other UK qualification types

If you choose either BTEC, NVQ or VRQ, the following suggested grades are available:
* Distinction
* Merit
* Pass
* Unclassified
* Not applicable

Suggesting UK vocational grades

## GCSE, AS level and A level subjects

When you start entering a subject for these qualifications, suggestions are provided:

Suggesting GCSE subjects

## GCSE, AS level and A level grades

When you start entering a grade for these qualifications, suggestions are provided:

Suggesting A level grades

## Maths, English and Science GCSE equivalents

~~I thought this work would touch on English/Maths/Science GCSE equivalents, but I’m not sure it does because:~~

* ~~There are many types of qualification for each of these subjects, some which may not be eligible; seems like if you selected English literature for example, or entered a grade D/5, we’d need a way of dealing with that scenario. There’s probably a much broader piece of design needed around equivalencies and minimum grades~~
* ~~We suggest that if you have a combined or triple GCSE in science, you enter all three grades as one grade; this is not a structured answer.~~


fofr:
(Aside: I've been wondering whether we should present "GCE O level" as "O level", at least in the GCSE section – people are understandably getting confused between GCE and GCSE.)

paulrobertlloyd:
@fofr Have changed ‘GCE O level’ to ‘O Level’ in the latest change.

PR #391, opened 2020-06-02
🌐 International qualifications

paulrobertlloyd:
* International degree flow
* International standard GCSE flow
* International academic or other relevant qualification flow (including English language qualification)

[Preview these changes here](https://apply-beta-prototype-pr-391.herokuapp.com/application/start)

### DIFF
--- a/app/views/application/gcse/naric.njk
+++ b/app/views/application/gcse/naric.njk
@@ -45,9 +45,7 @@
   {% endset -%}
 
   {% set noConditionalHtml %}
-    <h2 class="govuk-heading-s">If you studied for this qualification outside the UK, you will need to check that it is valid.</h2>
-    <p class="govuk-body">Contact your training provider to check their policy on non-UK qualifications.</p>
-    <p class="govuk-body">You can also register with <a href="https://getintoteaching.education.gov.uk">Get Into Teaching</a> to talk to a teacher training adviser. They’ll tell you if your qualification makes you eligible.</p>
+    <p class="govuk-body">Check that your qualification is valid for teacher training in the UK. You can ask your training provider or register with <a href="https://register.getintoteaching.education.gov.uk/register">Get Into Teaching</a> to talk to an adviser.</p>
   {% endset %}
 
   {{ govukRadios({
--- a/app/views/application/gcse/naric.njk
+++ b/app/views/application/gcse/naric.njk
@@ -18,7 +18,7 @@
       hint: {
         text: "For example, ‘My qualification is equivalent to a UK GCSE. I have a NARIC statement of comparability, UK NARIC reference 0000000000’"
       }
-    } | decorateApplicationAttributes(["gcse", id, "naric-details"])) }}
+    } | decorateApplicationAttributes(["gcse", id, "naric-reference"])) }}
   {% endset -%}
 
   {% set noConditionalHtml %}
--- a/app/views/application/gcse/naric.njk
+++ b/app/views/application/gcse/naric.njk
@@ -1,7 +1,7 @@
 {% extends "_form.njk" %}
 
 {% set subject = (id | capitalize) if id == "english" else id %}
-{% set title = "Do you know what UK qualification your " + subject + " qualification is equivalent to?" %}
+{% set title = "Do you have a NARIC statement of comparability for your " + subject + " qualification?" %}
 
 {% block pageNavigation %}
   {{ govukBackLink({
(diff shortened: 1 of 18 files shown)


app/views/application/gcse/naric.njk
```diff
@@ -13,18 +13,41 @@
...
   {% endset -%}
 
   {% set noConditionalHtml %}
-    <p class="govuk-body"><strong>If you studied for this qualification outside the UK, you will need to check that it is valid.</strong></p>
-    <p class="govuk-body">Contact your chosen training provider to check their policy on non-UK qualifications.</p>
-    <p class="govuk-body">Register at GOV.UK’s <a href="https://getintoteaching.education.gov.uk">Get into Teaching</a>. Registration will give you access to phone and email support from a teacher training adviser (TTA). Your TTA can tell you if your qualification makes you eligible for teacher training and how it corresponds to a UK GCSE.</p>
+    <h2 class="govuk-heading-s">If you studied for this qualification outside the UK, you will need to check that it is valid.</h2>
+    <p class="govuk-body">Contact your training provider to check their policy on non-UK qualifications.</p>
+    <p class="govuk-body">You can also register with <a href="https://getintoteaching.education.gov.uk">Get into Teaching</a>to talk to a teacher training adviser. They’ll tell you if your qualification makes you eligible.</p>
```
fofr, line 50:
```suggestion
    <p class="govuk-body">You can also register with <a href="https://getintoteaching.education.gov.uk">Get into Teaching</a> to talk to a teacher training adviser. They’ll tell you if your qualification makes you eligible.</p>
```

paulrobertlloyd:
> Should the degree subject be written in English by the candidate?

Good question; we did say this should be the case before.

> The steps in the degree journey say "save and continue", are all the answers saved as you go, or are they only persisted at the end? If it's the latter then they should just say "continue".

This old chestnut! We should be consistent with whatever we do elsewhere (we got pulled up on inconsistent button labels in the GDS content review)

> Does NARIC compatibility need a "I don't know" option, or some onward guidance about what NARIC is for people who select No?

Hmmm, possibly.

"Select the comparable UK degree" – is that an exhaustive list of qualifications, does it need an Other option?

Need to find out. Looking at the NARIC’s sample statement, does look as although these come from [this predefined list](https://www.naric.org.uk/naric/Code%20of%20Practice/Band%20Framework.aspx).

"As shown on your statement" - do we need to suggest where on the statement it is?

[Example statement](https://www.naric.org.uk/naric/Individuals/Compare%20Qualifications/documents/SoC.pdf). Maybe we say ‘found under the comparison subheading’ or similar?

> Might want to add PhD to "Doctor of Philosophy degree" as the commonly used term

I’m using the exact words as they appear on the statement, going by [this predefined list](https://www.naric.org.uk/naric/Code%20of%20Practice/Band%20Framework.aspx).

> The comparable list for undergraduate degrees include postgraduate equivalents, is that right?
Selecting "No" to the NARIC choice isn't shown in the summary, meaning, I can't go back and change my answer

Only for subsequent degrees; the first you degree you enter goes only up to masters (to account for integrated masters), i.e. the same logic we’re using elsewhere.

paulrobertlloyd:
> There's nowhere to put a grade for my international equivalent

Not sure how useful this is; the NARIC statement says if equivalent to ’GCSE (grades A*-C / 9-4)’, which is all that providers need to know. A grade value may be almost meaningless. Maybe.

When selecting No to NARIC, it also doesn't show the answer to that specific question (but shows comparable/NARIC number as not entered)

paulrobertlloyd:
> The steps in the degree journey say "save and continue", are all the answers saved as you go, or are they only persisted at the end? If it's the latter then they should just say "continue".

It appears individual steps are saved when adding a degree, so we should stick with ‘Save and continue’.

EmmaFrith:
1. 

on this page: https://apply-beta-prototype-pr-391.herokuapp.com/application/KVC11/degree/KVVSA?
say you’re an international student, as you see ‘you need an undergraduate degree…’
is this clear do you think? what about just ‘degree’?

2. 

On this page: https://apply-beta-prototype-pr-391.herokuapp.com/application/KVC11/degree/L6OVQ/institution?referrer=%2Fapplication%2FKVC11%2Fdegree%2Freview. There was something in the UR about ‘awarding’ being confusing
what about: Which institution did you get your degree from?
see UR - page 14: 
https://docs.google.com/presentation/d/1HLxjU9sidIpKRupV_9g8yVFLOHWZNBdEG_k4X4QH0Mk/edit#slide=id.g77ff765c1e_1_315

3. 

on this page: https://apply-beta-prototype-pr-391.herokuapp.com/application/KVC11/degree/LDYQC/naric - should we link to Naric?
A [NARIC statement of comparability](link) shows how your qualifications compare to UK qualifications.

4. 

did we decide to just move mention of ‘grades’ from the international quals section?
because they vary too much?

paulrobertlloyd:
> Say you’re an international student, and you see ‘you need an undergraduate degree…’
is this clear do you think? what about just ‘degree’?

How might using simply ‘degree’ work in the context of further degrees? We say: “You need an undergraduate degree to apply for teacher training. If you have further degrees, you can add them after.”

EmmaFrith:
Can we streamline to content under 'no' on the NARIC page to the following: 

Check that your qualification is valid for teacher training in the UK. You can contact your training provider or register with [Get into Teaching](link) to talk to an adviser for advice. 

---

Same advice but fewer words. We don't need to say 'if you got this qual outside UK' - they've already specified that it's an international qual.

paulrobertlloyd:
Merging in these changes, will create separate branch for further changes and refinements.

PR #393, opened 2020-06-05
Consistent capitalisation for Get Into Teaching

paulrobertlloyd:


paulrobertlloyd:
Going to do the same on production, FYI.

PR #395, opened 2020-06-11
Add English Language qualification/certification

paulrobertlloyd:
If you select a nationality in personal details of ‘Other’ or ‘Multiple’, a new ‘English language qualifications’ menu item appears under qualifications:

Screenshot 2020-06-11 at 19 39 08

There are three types of qualification you can enter; IELTS, TOEFL, and other (which represents the two main tests, for which we can ask more targeted answers, and a catch all for any other type of qualification:

Screenshot 2020-06-11 at 19 39 21

IELTS:

Screenshot 2020-06-11 at 19 39 36

TOEFL:

Screenshot 2020-06-11 at 19 40 00

Other:

Screenshot 2020-06-11 at 19 41 40

You can add multiple English Language qualifications, but we don’t optimise for it (as one certificate should be enough):

Screenshot 2020-06-11 at 19 46 21


adamsilver:
This is nice @paulrobertlloyd. Couple bits:

1. Should the H1s also be unabbreviated?
2. Can you make the 3 different score fields even smaller in width?

paulrobertlloyd:
* Made grade width 4 for IELTS and TOEFL, left it as 10 for other.
* Think we needn't expand abbreviation, as I believe IELTS and TOEFL well known abbreviation (probably more so than their full names).

PR #412, opened 2020-09-25
Simpler feedback form

paulrobertlloyd:
Replace multiple-step questionnaire with simpler feedback form that uses GDS pattern.

your-feedback


paulrobertlloyd:
@EmmaFrith First question probably needs updating, as unlikely to have completed our application in a day!

PR #418, opened 2020-10-22
Retry reference request

paulrobertlloyd:
(Addresses https://github.com/DFE-Digital/apply-for-teacher-training/pull/3205)

Remove ‘Change’ link next to email, and instead add a ‘Retry request’ action to the summary card header:

Screenshot 2020-10-22 at 17 23 46

This takes you to a page where you can resend request, and change email (repopulated with previous address used):

Screenshot 2020-10-22 at 17 23 59

Summary card status is changed to ‘Awaiting response’ and event added that says ‘Request sent to [new email address]’:

Screenshot 2020-10-22 at 17 27 09

### DIFF
--- a/app/routes/application/references.js
+++ b/app/routes/application/references.js
@@ -127,6 +127,15 @@ module.exports = router => {
         req.flash('success', `Reminder sent to ${applicationData.references[id].name}`)
       }
 
+      if (action === 'retry') {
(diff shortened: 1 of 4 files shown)


gregknight1 (commented):
Looks good from a content perspective, Paul.

PR #419, opened 2020-10-22
Feedback prompt

paulrobertlloyd:
First pass at a prompt to get feedback from candidates mid-service, based on [the feedback component used on GOV.UK](https://components.publishing.service.gov.uk/component-guide/feedback).

We need to adapt this given different context and what type of feedback we’d like to collect, and given it exists on a form, not a page of guidance.

JavaScript is just about passable (may have involved a bit of copy and paste from somebody’s CodePen ☺️), so needs a proper accessibility review to ensure attributes are being set correctly and focus is being delegated to the right elements.

Also need to consider state of component once feedback form has been submitted.

### Prompt question

feedback-question

### ‘Yes’ response

feedback-yes

### ‘No’ response

feedback-no


adamsilver:
What happens without JavaScript?

Can we find out about these ghost buttons and what other things they tried and failed along with they why etc?

Can we use big checkboxes and radio buttons? Once the form is there, we're not stuck for space or wanting this to be less prominent (I think).

adamsilver:
Content wise:

We go from ‘Can you complete this section?’ to ‘Help us improve this service’ (and a button that says ‘Submit feedback’) to ‘Thank you for your feedback’.

Is this about feedback or getting help with your application to proceed?

gregknight1:
I agree with Adam that the original question 'Can you complete this section?' infers that we might be about to help out with their application rather than ask for feedback.

We could alter the focus of the question from the applicant to the page. For example, 'Is there an issue with this page/section?' - the subsequent questions would have to alter slightly. We could even mention feedback explicitly in that question too. 

I'm not fully up to speed with this yet, and have other questions - I'll pick up with Paul.

PR #421, opened 2020-10-28
Remove references to needing 5 years of work history

paulrobertlloyd:
* Removes references to needing 5 years work history
* Simplifies question when starting work history section to yes/no (previous 4 options were always a bit odd)
* Updates guidance on missing work history page
* Removes ‘worked with children’ tick from summary card heading, adds this as a row with a change link instead
* Updates ordering of job summary card to match the ordering in the question
* …and yet another difference between work history and volunteering sections disappears. The two will be united, one day.

Also tidies up and refactors a few bits of the prototype around this section.

work-history

work-history-missing


gregknight1 (approved):
Having 2 radios is much clearer.

PR #422, opened 2020-10-29
Refactor feedback component, with focus on a11y

paulrobertlloyd:
Based on feedback, updated design that’s not as prominent as one used on GOV.UK (which is seen at the end of long pages of content), and improved implementation:

* Use `aside` to indicate form is not part of main flow
* Use `template` to hide parts of the design not needed if there is no JS
* Visually hide main question, rather than remove it from accessibility tree
* Only use `aria-controls`/`aria-expanded` logic for toggling of form(s)
* Over-engineer component so that I can change content of feedback form for signed in/out/other variants 😅

### Initial prompt
feedback-prompt

### Feedback form

feedback-form

With validation:

feedback-form-error

### Submitted feedback message (JS enabled)

feedback-success-js

### Feedback form (no-JS)

No close button is shown

feedback-form-no-js

### Submitted feedback page (no-JS)

feedback-success-no-js


app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
   GOVUK.Modules.Feedback = function () {
     this.start = function (element) {
```
adamsilver, line 10: I think the namespace should be ‘App’, not GOVUK

app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
   GOVUK.Modules.Feedback = function () {
```
adamsilver, line 9: I think the namespace should be ‘App’, not GOVUK

app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
...
+      let formContainer
+      let formCloseToggle
+      let formToggle
       let i
-      let target
-      for (i = 0; i < toggles.length; i = i + 1) {
-        toggleID = toggles[i].getAttribute('aria-controls')
-        toggleContent = document.getElementById(toggleID)
-        toggleContent.setAttribute('tabindex', '-1')
+      for (i = 0; i < formContainers.length; i = i + 1) {
```
adamsilver, line 49: I don't think the initialising should be within the component

app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
...
+        formContainer.setAttribute('tabindex', '-1')
+
+        // Add close button to each form before fieldset (and after any error summary)
+        const formFieldset = formContainer.querySelector('.govuk-fieldset')
+        formContainer.insertBefore(closeButtonClone, formFieldset)
+        formCloseToggle = formContainer.querySelector('.app-feedback__button--close')
       }
 
-      function toggle (ev) {
+      function toggleForm (ev) {
```
adamsilver, line 60: would use `e` as a convention - more common and shorter

app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
...
+            promptContainer.classList.add('govuk-visually-hidden')
           } else {
-            toggleContent.setAttribute('aria-hidden', 'true')
-            target.setAttribute('aria-expanded', 'false')
-            toggleQuestion.setAttribute('aria-hidden', 'false')
-            toggleQuestion.setAttribute('aria-expanded', 'true')
+            formContainer.setAttribute('aria-hidden', 'true')
+            formCloseToggle.setAttribute('aria-expanded', 'false')
+            formToggle.setAttribute('aria-expanded', 'false')
+            promptContainer.classList.remove('govuk-visually-hidden')
```
adamsilver, line 76: Are the polyfills available for this here? (I've asked on x-gov)

app/assets/javascripts/modules/feedback.js
```diff
@@ -9,41 +9,76 @@
...
           }
         }
       }
 
-      feedback.addEventListener('click', toggle, false)
+      feedback.addEventListener('click', toggleForm, false)
```
adamsilver, line 81: And here

adamsilver (changes requested):
Additionally, I wonder if the success state should be green/clearer

app/assets/javascripts/modules/feedback.js
paulrobertlloyd: Have added a test for `window.addEventListener`, `window.DOMTokenList` (for `classList`) and `window.HTMLTemplateElement` (for `<template>`s). If any of these tests fail, the simpler feedback form will be shown. I think that’s a good compromise given the widespread support for these methods.

paulrobertlloyd:
> Additionally, I wonder if the success state should be green/clearer

Have made success message green, see what you think.

app/assets/javascripts/modules/feedback.js
paulrobertlloyd: The `GOVUK.Modules` script appears to look for Modules with either `find` or `start` methods.

edwardhorsford:
The close button in secondary colour feels somewhat lost to me. I wonder if the action could be clearer / something else improved about it - blue to match the intiall buttons?

app/assets/javascripts/modules/feedback.js
adamsilver: Here's accordion using init?
https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/accordion/accordion.js

PR #426, opened 2020-11-06
Content update - references start

gregknight1:
Include content on SD Salaried courses.
Content to encourage candidates to ensure employers give full references.

@paulrobertlloyd Can you check the CTA on this page too please? Not sure if I've broken something.


paulrobertlloyd:
@gregknight1 Fixed CTA. Also moved this line to be last paragraph of first section of content on this page:

> You can add as many referees as you like to increase the chances of getting 2 references quickly.

This is so that the preceding three paragraphs that relate to the different types and when/why you should choose them are grouped together. What do you think?

PR #427, opened 2020-11-10
Content Request cancelled status

gregknight1:
Include content to explain why references are cancelled when 2 are received.


gregknight1:
@paulrobertlloyd Can you check this additional content to the `Request cancelled` status, please.

PR #431, opened 2020-11-18
Add .nvmrc

frankieroberto:
This allows to to type `nvm use` if using the node version manager, to switch to using the right version of node.


fofr:
Why only 10.*?

https://github.com/alphagov/govuk-prototype-kit/blob/master/.nvmrc

frankieroberto:
@fofr I was following the version specified in `package.json` which is currently `^10.0.0` (which I think is equivalent to `10.*`, but npm doesn't support the `^` operator I think).

PR #434, opened 2020-11-20
Iteration of the work history section

frankieroberto:
# Work history iteration

This iteration of the Work history section of the application process aims to address the following issues:

* The [initial flow](https://bat-design-history.netlify.app/apply-for-teacher-training/work-history/) suggested that you only needed to provide work history for the past 5 years, however providers have indicated that they need a complete work history for safeguarding purposes. This could significantly increase the burden on candidates with a long work history.
* Candidates are currently asked for *Skills and experience relevant to teaching you gained in this role* for every job, and spend a considerable amount of time writing this. However, providers may only quickly scan this, and answering this question for non-relevant roles such as working in a cafe may be difficult and stressful. In addition, validation errors suggest that when faced with this field, many candidates leave it blank, perhaps intending to come back and fill it in later.
* For part time roles, candidates are currently asked to give details of their working pattern. This may be difficult to remember or describe for variable-hours roles or jobs done a long time ago.
* Candidates who have always been in full time education, for example those in their final year of University who did not take any gap year, are currently asked to describe why they have been out of the workplace. This is unneccessary and could be made easier. Statistics show that around 20% of candidates are 20 or 21.
* Currently, unpaid experience is a separate section. However it could overlap with the work history section, if unpaid work explains a gap in paid employment. It can also overlap with the personal statement section, which asks for details of school experience or volunteering with children.

## Prototype

➡️  You can [view the proposed changes as a working prototype](https://apply-beta-pr-434.herokuapp.com). The password is the same as the regular prototype.

## Approach

Some of the issues seem to arise from the work history section have two distinct purposes: supporting a candidate’s positive qualities in the application, and avoiding potential harms via safeguarding. These conflict with each other when it comes to deciding how much detail is required, and how much effort candidates should spend on the section.

Providers have indicated that the primary purpose for asking for a complete work history is for safeguarding, and that more emphasis is placed on the personal statement when ranking candidates for offering places.

Therefore, this design proposes moving the work history section to be beneath the "Safeguarding" heading in the application. In addition, the fields are reduced to being a purely factual representation of work history, rather than requiring additional thought and writing.

The Unpaid experience section is removed, with the intent that significant unpaid jobs which help explain employment gaps (such as a 3 month volunteer role abroad) get added to the work history section, and other volunteering which helps demonstrate teaching suitability gets mentioned in the personal statement.

### Work history initial question

Currently, the inital question looks like this:

Screenshot 2020-11-23 at 15 08 12

The proposed design adds an option for candidates who have always been in full time education, to avoid them having to describe this in words:

Screenshot 2020-11-23 at 15 09 28

### Tell us why you’ve been out of the workplace question

This currently looks like this:

Screenshot 2020-11-23 at 15 16 03

The proposed design makes this a conditionally-revealed question on the initial page:

Screenshot 2020-11-23 at 15 17 14

The main advantage of this is to make it easier to switch between the 3 different answers. It may also make the question feel less daunting.

### Add a job

This page currently looks like this:

add-job-old

The proposed design is broadly similar, with the following differences:

* The *skills and experience* question has been dropped
* The full time vs part time question has some hint text, and the conditionally-revealed working pattern question has been dropped
* The end date question has been changed to ‘Are you still working in this job’, and a month and year is asked if ’No’ is selected.
* The start date question has been reworded to ‘When did you begin this job?’ for consistency with the end date question, and to better reflect that we are not asking for a full date.
* The ‘Do you want to add another job?’ question has been dropped (see below for rationale)

add-job-new

### Work history review

The page currently looks like this:

work-history-old

The proposed design looks like this:

work-history-new

The biggest difference is using a grid (table) view, instead of summary cards. The main reason for this is to make the history easier for applicants to scan to check for completeness, by having the dates of each job closer to each other in columns.

The additional text of “(part time)” and ”This role involved working with children” only appear when selected.

A “Remove” link appears within the Actions column – however another option might be moving this to within the edit form, as this is likely a rare action.

Where the candidate indicates they have always been in full time education, the following in shown:

work-history-full-time-education

### Change to flow

Currently, when adding jobs, candidates are asked if they would like to add another job, in which case they are taken to a blank form for a new job.

The design proposes changing this flow so that candidates are always taken back to the review page after adding a job. The rationale here is that, if adding lots of jobs as part of a long work history, it will be easier for candidates to keep track of which ones they’ve added so far if they return to the review page each time. However, this does come at the expense of an extra click.

## Implementation thoughts

Given that this is a significant change which is not backwards-compatible (as fields are being dropped), it could be very confusing for candidates who are currently mid application, or reviewing a previous application.

Therefore it may make sense for all existing applications to use the existing design, and for this design to only be used for new applications.

Likewise, there may need to be different designs for the provider interface, depending on whether the candidate used the old design or the new one.

## Thoughts for the future

By moving the work history section to be part of safeguarding, we make it more possible to make changes to who sees this and when in future. For instance, it could get reviewed by a different team from the rest of the application, or at a different stage in the process.


paulrobertlloyd:
This is great! Be interesting to see how it tests with candidates that need to enter many jobs. A few notes/questions:

### Start page

* Should probably update prototype so that the initial question isn’t shown after you’ve added your first entry.

### Add/edit role

* **Label sizes.** I note that you’ve opted to use the smaller label size. There was a long running debate about this, back in the day, and we ended up using the medium size. Maybe we can revisit this decision (alongside using large page titles instead of extra large ones), but for now, should probably be consistent with the sizing we’re using elsewhere in the application.
* **Full/part time.** Given most jobs will likely be in the past, should ‘Are you still working in this job?’ default to ‘No’ (one less click) or does that add visual complexity?

### Review

* **Timeline typography.** Wondering if roles titles should be emboldened. This would help visually separate each entry, and also help distinguish between role name and any commitment text (i.e. ‘Part time’)
* **Break guidance.** Do we need to say ‘Any gaps of longer than one month should be explained’ on the review page, if such gaps show up anyway?
* **Break options.** We currently offer 2 options where we show a break; explain break, or add a job. Was there a reason for only having the one option instead?
* **Break in history.** Would be good to see what a break looks like in the timeline

gregknight1:
Really good, Frankie.

It's good that you've made the start and end date questions consistent. And also that you funnel users for an end date which removes lots of words - great.

We could look at reviewing the guidance at the top of the review page so it's more active. For example:

Include all the jobs you've had as an adult, including unpaid roles.
Do not include part time jobs you had during full time education.
Explain any gaps of one month or longer.

The CTA after removing a period of work, I think, has some extra words in that shouldn't be there:

image

Probably should be: Yes, I'm sure

gregknight1 (approved):
Looks good.

frankieroberto:
@paulrobertlloyd thanks for your comments. I've changed/fixed all the things, apart from this:

> should ‘Are you still working in this job?’ default to ‘No’

I'm leaning towards not having a default for now, on the basis both that it feels weird to start the page with conditionally revealed content already revealed, plus having a default set for the radio button doesn't allow the user to leave the choice unset (maybe they're not yet sure if they're going to leave their job or not?) or they might not notice that it's already been answered.

However, let's see how it tests!

frankieroberto:
@gregknight1 thanks, I've made all those changed (and fixed the bug in the "remove" button text, opps). 👍

adamsilver:
A few thoughts:

1. Put the full time hint on the full time radio button directly
2. Concerned about the red for the remove link or the confirm button. See guidance or information I sent on Slack
3. Can we stop the action links from stacking
4. It may be hard for some users to spot the issue with the work gap – perhaps it needs to be called out clearly without being in the middle of the list which may only (semi) work okay for sighted users
5. Did you consider asking the user whether they wanted to add another entry on a separate page?

app/views/application/work-history-2/review.njk
```diff
@@ -17,9 +17,9 @@
 {% block primary %}
 
   <div style="max-width: 650px;">
```
paulrobertlloyd, line 19: 👀

PR #435, opened 2020-11-24
Update notification banners to use recommended pattern

paulrobertlloyd:
The design system now has [a recommendation for notification banners](https://design-system.service.gov.uk/components/notification-banner/) – hooray!

In most cases in looks as though we can do a straight swap of what we used previously with this new design. However, there are a few issues:

* ‘Important’ doesn’t seem like the right heading for the informational banner shown when you can apply again. Looking for suggested alternatives for this heading:

  Screenshot 2020-11-24 at 16 09 14
  
* We probably want to use the blue important style, not the green success variation for the ‘Your new application is ready for editing’ message.

Also…

* The section incomplete messages we show on the application review page previously used the old banner style. This was changed in production to something more bespoke, so I have updated the prototype to reflect this change.

  However… production still inherits and modifies styles from the older banner design, so we need to create a new component so we can remove this dependency. Have gone with `app-summary-text`. Coloured blue by default, it also has an error variation that aligns with error message style used on forms. Could alternatively use `govuk-inset-text` component and modify it, but given we always show a heading and an action and need to style these, thought best to create something more bespoke.

  Screenshot 2020-11-24 at 15 25 09
  Screenshot 2020-11-24 at 15 25 32


frankieroberto (commented):
Nice work on adoption the new banners so fast! 😄 

I wonder whether "Do you want to apply again?" really should be a notification banner or not... As you say, "Important" feels a bit wrong, which perhaps indicates it’s the wrong component? Is it also scoped only to the application page, or would it appear if you also clicked through to view the old application?

I was surprised that this page doesn't use the banner? This one _does_ feel important, given that there’s a deadline?

Screenshot 2020-11-25 at 11 54 40

"summary text" feels a bit too generic to me, but I don't have any immediate better suggestions. Although perhaps @adamsilver's suggestion of a modifier of the Inset text could work?

paulrobertlloyd:
> I was surprised that this page doesn't use the banner? 

Oooh, good point. This message is definitely banner worthy. We should look at that separately as part of dashboard review. Apply again messaging is part of that too. Question is whether to do something in the meantime? Not like there’s a great rush to add the new banners…

> Although perhaps @adamsilver's suggestion of a modifier of the Inset text could work?

Updated to do this. Actually turned out to require less code, so most definitely the right approach.

PR #437, opened 2020-11-30
Update Start page to include free NARIC statement

gregknight1:
@frankieroberto @paulrobertlloyd - could either or both of you cast your eyes over this additional content for the start page before I request the change with GDS, please?

NARIC statements give people who have qualifications from overseas their comparable UK qualifications/grades. 

We already link to the NARIC website on our start page. We don't tell candidates how much they cost or that they can get free statements through Get Into Teaching.

Include this content and move the 'If you studied overseas content' under the 'Get advice and support' H2

I think moving the content makes it flow better because of where we refer to Get Into Teaching (first referred to under Get advice and support), rather than just tagging it on the end of the current 'If you studied overseas' content.

Current state:

image

Proposed state:

image

### DIFF
--- a/app/views/index.njk
+++ b/app/views/index.njk
@@ -60,14 +60,16 @@
     <li class="govuk-body">grade 4 (C) or above in English and maths GCSEs (or equivalent)</li>
   </ul>
 
-  <h3 class="govuk-heading-s">If you studied overseas</h3>
-  <p class="govuk-body">You can <a href="https://www.naric.org.uk/Qualifications/SOC/Default.aspx">get a statement from the National Recognition Information Centre for the UK (UK NARIC)</a> that shows how your qualifications compare to UK qualifications. Some training providers may ask to see it as part of your application.</p>
-
   <h2 class="govuk-heading-m">Get advice and support</h2>
   <p class="govuk-body">You can <a href="https://getintoteaching.education.gov.uk/">register with Get Into Teaching</a>, the Department for Education’s free support and advice service.</p>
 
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
Looks good to me!

frankieroberto:
@gregknight1 agree with the reordering.

I didn't know anything about these NARIC statements until reading this, but it took me a couple of readings of this guidance to realise that "statement of comparability" was the same thing as the statement referred to in the previous paragraph. Is there any way to make this clearer perhaps, or am I just being slow?

gregknight1:
@frankieroberto I did hesitate when writing that. It's a perfectly valid point - they are called statements of comparability but a candidate on their first application wouldn't necessarily know that at this point.

I was hanging on to the theory that we've said 'statement' and 'compare' in the previous para so hoping it wasn't too much of a leap. Glad you didn't know about them before as that tells me it is too much of a leap.

Maybe: 'You may be able to get a statement that compares your qualifications for free'

frankieroberto:
@gregknight1 I think that's better..!

Are we able to give any more details on "may be able to" get it for free? Feels tantalising, but a bit vague... 😄

gregknight1:
@frankieroberto Yep, I also considered this - but to explain the conditions would require quite a bit more content here. I had to phone the number to find out as the conditions on whether you can or can't get a free statement aren't published, so we can't link to them.

The GiT advisor told me that most people would get a free statement if they called but there are a few nuanced exceptions according to where/when they studied.

Felt best to keep it simple but with that simplicity comes a bit of vagueness.

paulrobertlloyd:
@gregknight1 @frankieroberto That tripped me up initially as well, and was halfway through suggesting a change, but then ended up where Greg landed.

PR #438, opened 2020-12-01
Content: referee cannot save and return

gregknight1:
For MVP we need to tell referees that the reference has to be done in one sitting.

Copy on top of the reference submission page to inform the referee.

@paulrobertlloyd Could you check that this is ok, please? Does it need the warning component or any other to differentiate it from the reference advice text, do you think?

Having considered other places for this copy, I think this probably the best. If we put it in the email it could put referees off at an early stage and they might not come back for a while. If they have this copy here I feel they'd be more likely to carry on.

### DIFF
--- a/app/views/reference/comments.njk
+++ b/app/views/reference/comments.njk
@@ -18,6 +18,7 @@
 {% endblock %}
 
 {% block primary %}
+  <p class="govuk-body"> You have to complete the reference in one sitting. You cannot save it and come back to it later.</p>
   <p class="govuk-body">When giving your reference try to cover the following, but do not worry if there are areas you cannot comment on:</p>
   <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
 {% if type == "academic" %}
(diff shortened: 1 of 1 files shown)


app/views/reference/comments.njk
```diff
@@ -18,6 +18,7 @@
 {% endblock %}
 
 {% block primary %}
+  <p class="govuk-body"> You have to complete the reference in one sitting. You cannot save it and come back to it later.</p>
```
paulrobertlloyd, line 21: Is saying ‘you have’ a bit too direct, perhaps going again the softer tone we are aiming for? Especially as it’s then followed by ‘cannot’.

Probably want to stay away from the warning component; that tends to be used in very rare circumstances. If this continues to prove a problem, maybe we can look at other approaches? 

```suggestion
  <p class="govuk-body">Complete your reference in one sitting. You cannot save it and come back to it later.</p>
```

gregknight1:
@paulrobertlloyd Thanks - I think that's better. I was toying round with trying to soften it by saying 'You have to', 'You need to' etc - but you're right, just makes it harsher.

PR #442, opened 2020-12-03
Collect multi-line international addresses

paulrobertlloyd:
Will investigate other options such as breaking multiline input from a textarea, but I think still that’s ripe with issues converting to a frame that’s suitable for sending to SRSs.

So, alternatively, we can repurpose the inputs we use for UK addresses, but adjust the input labels to show international alternatives. We could customise labels to specific countries, but not sure how much effort that’s worth given current number of international students.

I think the bigger issue is formatting of addresses when previewed; this is where they tend to differ.

| UK address | International address |
| - | - |
| uk-address | international-address |

### DIFF
--- a/app/data/countries.js
+++ b/app/data/countries.js
@@ -221,9 +221,6 @@ module.exports = [{
   value: 'Dubai',
   text: 'Dubai'
 }, {
-  value: 'East Germany',
-  text: 'East Germany'
-}, {
   value: 'East Timor',
   text: 'East Timor'
 }, {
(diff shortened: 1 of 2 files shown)


frankieroberto:
Do we also need candidates to enter the country name for international addresses?

Does it matter that order won't match the typical order of addresses for some countries - eg I think some countries put the postcode before the region, so they might end up with their address in the "wrong" order?

I'd be tempted to try:

Address line 1
Address line 2
Address line 3 (optional)
Address line 4 (optional)
Address line 5 (optional)

😂

PR #443, opened 2020-12-10
Update prototype to use files from 9.11.2

frankieroberto:
These file changes are a straight copy from the Prototype Kit 9.11.2


gulp/sass.js
```diff
@@ -28,3 +28,28 @@ gulp.task('sass', function () {
...
+      outputStyle: 'expanded',
+      includePaths: [
+        'node_modules/govuk_frontend_toolkit/stylesheets',
+        'node_modules/govuk-elements-sass/public/sass',
+        'node_modules/govuk_template_jinja/assets/stylesheets'
+      ]
+    }).on('error', sass.logError))
+    .pipe(sourcemaps.write())
+    .pipe(gulp.dest(config.paths.public + '/v6/stylesheets/'))
+})
```
paulrobertlloyd, line 55: Don’t need these tasks as we no longer have the documentation folder, or need v6 support

gulp/watch.js
```diff
@@ -16,3 +16,14 @@ gulp.task('watch-assets', function () {
...
+// Backward compatibility with Elements
+
+gulp.task('watch-sass-v6', function () {
+  return gulp.watch(config.paths.v6Assets + 'sass/**', { cwd: './' }, gulp.task('sass-v6'))
+})
+
+gulp.task('watch-assets-v6', function () {
+  return gulp.watch([config.paths.v6Assets + 'images/**',
+    config.paths.v6Assets + 'javascripts/**'], { cwd: './' }, gulp.task('copy-assets-v6'))
+})
```
paulrobertlloyd, line 29: Don’t need these tasks as we don't need to support v6

gulpfile.js
```diff
@@ -21,12 +21,18 @@ gulp.task('generate-assets', gulp.series(
   'sass-extensions',
   gulp.parallel(
     'sass',
-    'copy-assets'
+    'copy-assets',
+    'sass-documentation',
+    'copy-assets-documentation',
+    'sass-v6',
+    'copy-assets-v6'
```
paulrobertlloyd, line 28: Ditto

gulpfile.js
```diff
@@ -21,12 +21,18 @@ gulp.task('generate-assets', gulp.series(
...
   )
 ))
 gulp.task('watch', gulp.parallel(
   'watch-sass',
-  'watch-assets'
+  'watch-assets',
+  'watch-sass-v6',
+  'watch-assets-v6'
```
paulrobertlloyd, line 35: Ditto

listen-on-port.js
```diff
@@ -8,8 +8,8 @@ const config = require('./app/config.js')
 const utils = require('./lib/utils.js')
 
 // Set up configuration variables
-const useBrowserSync = config.useBrowserSync.toLowerCase()
-const env = (process.env.NODE_ENV || 'development').toLowerCase()
+var useBrowserSync = config.useBrowserSync.toLowerCase()
+var env = (process.env.NODE_ENV || 'development').toLowerCase()
```
paulrobertlloyd, line 12: This shouldn't change anything, and using `var, when linting, Standard complains.

listen-on-port.js
```diff
@@ -22,7 +22,7 @@ utils.findAvailablePort(server, function (port) {
         port: port,
         ui: false,
         files: ['public/**/*.*', 'app/views/**/*.*'],
-        ghostmode: false,
+        ghostMode: false,
```
paulrobertlloyd, line 25: I suspect this might be our culprit!

package.json
```diff
@@ -14,47 +14,50 @@
...
+    "notifications-node-client": "^4.7.2",
+    "nunjucks": "^3.2.1",
     "portscanner": "^2.1.1",
-    "prompt": "^1.0.0",
-    "querystring": "^0.2.0",
     "require-dir": "^1.0.0",
-    "sync-request": "^6.1.0"
+    "sync-request": "^6.0.0",
+    "universal-analytics": "^0.4.16",
+    "uuid": "^7.0.3"
```
paulrobertlloyd, line 49: Not sure we need to downgrade dependencies… `humanize-duration`; I think that this is still being used for a Nunjucks filter, and used for work history breaks. `inquirer` is something that was added with a more recent version of the kit, might be worth investigating what changed to require this.

package.json
```diff
@@ -14,47 +14,50 @@
...
   },
   "devDependencies": {
-    "jest": "^26.6.3",
-    "standard": "^16.0.2",
-    "supertest": "^6.0.1"
+    "glob": "^7.1.4",
+    "jest": "^25.2.7",
+    "node-sass": "^4.14.1",
+    "standard": "^14.3.3",
+    "supertest": "^4.0.2"
```
paulrobertlloyd, line 61: Again, not sure need to downgrade.

server.js
```diff
@@ -1,15 +1,15 @@
 // Core dependencies
+const fs = require('fs')
 const path = require('path')
 
 // NPM dependencies
+const bodyParser = require('body-parser')
 const dotenv = require('dotenv')
 const express = require('express')
-const flash = require('connect-flash')
```
paulrobertlloyd, line 7: Need this!

server.js
```diff
@@ -1,15 +1,15 @@
...
 const nunjucks = require('nunjucks')
 const sessionInCookie = require('client-sessions')
 const sessionInMemory = require('express-session')
 const cookieParser = require('cookie-parser')
-const addNunjucksFiltersWithAppContext = require('./app/utils/filters-with-app-context')
```
paulrobertlloyd, line 12: Need this too!

paulrobertlloyd (changes requested):
Going to stop reviewing. This change will likely break a lot of things. If it fixes the ports issue we should narrow down what specific change is causing it.

frankieroberto:
@paulrobertlloyd ah, I hadn't realised that you'd deliberately edited these files. I've always tried to not-edit any of these files, to make upgrading a bit easier... (although it's still a pain).

server.js
fofr: This is a big one. This is something I think we should push upstream to the prototype kit, it's so useful.

frankieroberto:
Closing this as it's clear that my naive just-copy-and-paste won't work, so we'll need to cherry-pick any updates.

PR #445, opened 2021-01-13
Further iteration on work history

frankieroberto:
Changes from version tested with:

* Re-instate "Work experience" section in review page, and reinstate unpaid experience. Research showed that this information was _not_ just used for safeguarding after all.
* Move guidance that all jobs since 18 need to be added from the top of the review page to the top of the form, to make it more visible (not all users saw this in testing).
* Add explanation of how the information is used to the top of the review page (content tbc), and advising candidates to mention relevant experience in their personal statement.
* Move employer before job title – as this is what people remember first, and 1 participant entered the employer name into job title field repeatedly.
* Add hint text to job title to explain that only last role is needed, if the user was promoted within the organisation.
* Add hint text to explain that candidates only need to enter an estiamted month and year if they can't otherwise remember it.
* Add checkbox for candidates to indicate that they couldn’t remember the month or year they started or finished a job.
* Reworded the "Did this job involve working with children?" to be "Did you use skills relevant to teaching in this role?"


frankieroberto:
Here's an idea I'm toying with for making the breaks in work history more visible (as some users in testing missed the mid-table prompts). The basic idea was to re-use the error summary, but make it blue to indicate it's not completely mandatory. The blue links are anchor links to the mid-table prompts.

Screenshot 2021-01-14 at 16 49 45

paulrobertlloyd:
* Hint text for *Name of employer* (If you worked for yourself, enter ‘self-employed’) – should this be ‘Self-employed’ (capital S)?
* ‘Save and continue’ button on review page doesn’t work; should this take you to the application page?
* Should the work history timeline sit within a box, to match how we play back results in each of the other sections (and on the review page)?
* The presentation of work history without a table I think works better (certainly on mobile), but wonder if the blue border for a break should extend to the entire height of its space; having a margin above and below it makes the timeline a bit ‘noisier’.
* The application review page needs updating to reflect the latest changes; work history no longer sits under safeguarding section, and no longer uses table to present data
* Volunteering section now uses a different format for displaying a history. Maybe that’s fine, as the content provided is more discursive, whereas work history is just dates.
* Should we replace the previous work history design with this new one, save having to manage the two different versions?

paulrobertlloyd:
Oh, and the JS could do with a lil’ linting 😉

frankieroberto:
@paulrobertlloyd so, even though it looks a bit odd perhaps, @gregknight1 and I agreed that "self-employed" should be lower case in the hint text example, as it's not a 'proper noun'.

Perhaps ideally self-employment would be a separate question. Something like:

```
Were you employed or self-employed?

() Employed
|  Name of employer
|  [          ]
() Self-employed
```

...but that seems unnecessary for now - although perhaps we could revisit this in future if self-employment proves to be tricky and/or common.

gregknight1:
Yes, it should be self-employed here. 

Are we not moving the 'add another' button for work history like we are for courses, references and qualifications - it will impact the guidance like it has in those other sections? @frankieroberto @paulrobertlloyd

paulrobertlloyd:
> Are we not moving the 'add another' button for work history like we are for courses, references and qualifications - it will impact the guidance like it has in those other sections?

I think we should move this up, to be consistent with how we are now doing it on other review pages.

gregknight1:
Looking at the guidance for the indented text in terms of moving the 'add another' button up.  I still think it's needed as the applicant is taken straight into the form and doesn't necessarily see the guidance again if adding multiple jobs in one sitting.

We could make it shorter and more succinct, though. Something like:

_**Enter all the jobs you've had since you left school.

Explain any gaps in your work history. For example, raising children, unemployment or illness.**_

frankieroberto:
@paulrobertlloyd ok, I think I've addressed most of these issues, and I've removed the 'old' flow now, so the new flow no longer has a `-2` suffix. 

> Should the work history timeline sit within a box, to match how we play back results in each of the other sections (and on the review page)?

Hmm, I think this is a bit unnecessary on the work history review page, and clashes a bit with the horizontal rules between the jobs, but perhaps it is needed on the overall application review page, where all the other sections sit in a bordered box?! Needs a bit more thought.

> The presentation of work history without a table I think works better (certainly on mobile), but wonder if the blue border for a break should extend to the entire height of its space; having a margin above and below it makes the timeline a bit ‘noisier’.

Here's a quick try of this:

no-margin

vs

margin

– I'm not sure it makes it stand out any more, and maybe clashes a bit with the grey horizontal line? Don't have a strong opinion on this though.

paulrobertlloyd:
This is looking great! Thanks for persisting on this @frankieroberto. I know it was a bit of a pain, but I really like where we’ve ended up; it‘s a huge improvement on what we have currently. :shipit:

PR #448, opened 2021-01-25
Add additional application states to prototype

frankieroberto:
This updates the prototype to show more different states and combinations of states in the post-submission phase.  These mostly now reflect the current production design (although the apply again prompt is still a bit different).

There is some information deliberately missing, as it hasn't yet been added to the live service:

* offer conditions (once accepted)
* reasons given by the provider for withdrawing an offer
* reasons given by a candidate for declining an offer or withdrawing an application
* ability to see previous applications and outcomes (if applying again)

This PR is mainly just laying groundwork for a rethink and review of these sections.

### DIFF
--- a/app/data/application-single-choice.js
+++ b/app/data/application-single-choice.js
@@ -1,0 +1,14 @@
+// Copies dummy application and tweaks some fields
+const application = JSON.parse(JSON.stringify(require('./application')))
+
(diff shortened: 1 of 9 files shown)


gregknight1 (approved):
Really helpful, thanks Frankie.

PR #450, opened 2021-01-28
Dashboard redesign

frankieroberto:
This updates the post-submission view that candidates see after logging in (previously known as "Application dashboard").


paulrobertlloyd:
Thoughts (some of which might be easier to go through in person):

## Awaiting decisions

* Why do we show contact details for a provider if there’s only 1 application, but not if there’s more?

## Interviewing

* Margin is missing spacing below the status tag when only 1 application

## Offer received

* Probably a bug in the prototype, but seems like you can’t respond to an offer in the _One offer received, still awaiting one provider decision_ state?
* Inconsistent spacing between various elements in that state too.
* The layout with 1 offer is again a bit hard to parse with lots going on, layout and type wise.

## Unsuccessful

* I wonder if having 2 different dates in this text makes thing more confusing:

  > You can apply again for courses starting in September 2021. You have until 5 August 2021 to apply again.

## Accepted offer, pending conditions

* This state shows an accepted offer, with 2 other choices awaiting decision. Once you’ve accepted an offer, all other choices get declined, so this state wouldn’t be possible.

## Respond to offers

* It’s entirely unclear to me that there are two separate application choices on this page; it’s a wall of text – and likely only to be more so if there are multiple conditions of unknown length. I think we need to delineate the different offers on this page.

## Overall

* I like the new withdraw application flow, and that it covers all options (all/one/some).
* I like that we’re taking into consideration various new states.
* Still concerned about having 2 different layouts; one for 1 choice, another for multiple choices. If it could be proven that having a different layout benefited the content, then I could go with it, but right now, the layout for 1 course doesn’t hang together well.
* I wonder if maybe we could look at the different states of a single application choice, and see how/if they could be designed such that it wouldn’t matter if there was 1, 2 or 3 choices on a page. Maybe that’s rows, maybe its cards, maybe it’s something else?
* I wonder if it’s a problem that the main call to action is separate from the course choice it relates to below?
* Can’t say I’m a fan of the inset text area with large text and blue border; feels like its a completely new and unrelated pattern to what we’re using elsewhere. Why not use the new banner component? Seems like this is a situation for which that component was designed for? If not, maybe the boxout pattern we’re now using on Find?

> Is ‘Apply Again’ the right terminology - does it suggest a whole new application for next year?

It was always intended that ‘apply again’ (lowercase) was a lightweight choice. You can always apply again – this cycle or next cycle – just that the options change depending on when you do so (if you apply again this cycle you have one choice, if you apply again next year, you have three choices). I’d be very tempted not to over complicate things beyond that.

gregknight1:
> Thoughts (some of which might be easier to go through in person):
> 
> ## Awaiting decisions
> * Why do we show contact details for a provider if there’s only 1 application, but not if there’s more?
> 
> ## Interviewing
> * Margin is missing spacing below the status tag when only 1 application
> 
> ## Offer received
> * Probably a bug in the prototype, but seems like you can’t respond to an offer in the _One offer received, still awaiting one provider decision_ state?
> * Inconsistent spacing between various elements in that state too.
> 
> ## Unsuccessful
> * I wonder if having 2 different dates in this text makes thing more confusing:
>   > You can apply again for courses starting in September 2021. You have until 5 August 2021 to apply again.
> 
> ## Accepted offer, pending conditions
> * This state shows an accepted offer, with 2 other choices awaiting decision. Once you’ve accepted an offer, all other choices get declined, so this state wouldn’t be possible.
> 
> ## Respond to offers
> * It’s unclear to me that there are two separate application choices on this page; it’s a wall of text – and likely only to be more so if there are multiple conditions of unknown length. I think we need to delineate the different offers on this page.
> 
> ## Overall
> * I like the new withdraw application flow, and that it covers all options (all/one/some).
> * I like that we’re taking into consideration various new states.
> * Still concerned about having 2 different layouts; one for 1 choice, another for multiple choices. If it could be proven that having a different layout benefited the content, then I could go with it, but right now, the layout for 1 course doesn’t hang together well.
> * I wonder if maybe we could look at the different states of a single application choice, and see how/if they could be designed such that it wouldn’t matter if there was 1, 2 or 3 choices on a page. Maybe that’s rows, maybe its cards, maybe it’s something else?
> * I wonder if it’s a problem that the main call to action is separate from the course choice it relates to below?
> * Can’t say I’m a fan of the inset text area with large text and blue border; feels like its a completely new and unrelated pattern to what we’re using elsewhere. Why not use the new banner component? Seems like this is a situation for which that component was designed for? If not, maybe the boxout pattern we’re now using on Find?
> 
> > Is ‘Apply Again’ the right terminology - does it suggest a whole new application for next year?
> 
> It was always intended that ‘apply again’ (lowercase) was a lightweight choice. You can always apply again – this cycle or next cycle – just that the options change depending on when you do so (if you apply again this cycle you have one choice, if you apply again next year, you have three choices). I’d be very tempted not to over complicate things beyond that.

Great points, thanks @paulrobertlloyd - think you're right, probably best to discuss in person on Monday.

frankieroberto:
@paulrobertlloyd thanks for your review...!

> Why do we show contact details for a provider if there’s only 1 application, but not if there’s more?
Interviewing

This was an oversight, but also weren't sure how helpful it was to display these details, given that the candidate might have been given a more specific or personal contact. But perhaps we should try and include it still. At the very least, we could include it on the application detail view.

> Probably a bug in the prototype, but seems like you can’t respond to an offer in the One offer received, still awaiting one provider decision state?

This was sort-of a feature, in that we perhaps don't want candidates to accept an offer (or feel pressured into accepting an offer) before all providers have responded. They _could_ accept the offer by withdrawing from the courses still waiting a decision, but this is probably a bit hidden. The research seemed to show that some candidates would want to accept straight away if their preferred choice made an offer, so I think we probably do need a more obvious way of doing this. We could either show the call to action straight away (but might be too prominent), or add an action within the application view, or a link within Next steps (alongside withdraw). Perhaps we should include some content warning candidates that they don't need to accept until all responses are in?

> I wonder if having 2 different dates in this text makes thing more confusing:
> 
> You can apply again for courses starting in September 2021. You have until 5 August 2021 to apply again.

I think @gregknight1 will like rework this.

> Respond to offers
> 
> It’s unclear to me that there are two separate application choices on this page; it’s a wall of text – and likely only to be more so if there are multiple conditions of unknown length. I think we need to delineate the different offers on this page.

Yeah, this definitely needs some more thought.

> Can’t say I’m a fan of the inset text area with large text and blue border; feels like its a completely new and unrelated pattern to what we’re using elsewhere. Why not use the new banner component? Seems like this is a situation for which that component was designed for? If not, maybe the boxout pattern we’re now using on Find?

Let's keep looking at this. I did look at the [notification banner](https://design-system.service.gov.uk/components/notification-banner/) but it seems that it was designed to appear above the h1, containing information not directly relevant to the page below, rather than containing a call to action directly related to the page. I also slightly fear it's so prominent that it'll get ignored (banner blindness).

Tomcowan1:
Hi @frankieroberto thanks for putting the content in. Suggested changes:

For multiple courses:

- In "_Offer deferred_" change the second line to: _Before your offer is confirmed, there are some final conditions to meet._ Also, there aren't any conditions showing in this state... should there be?
- We talked about separate states for Candidates declining all offers, Candidate doesn't respond to offers in time, and Candidate withdrew from all offers, but they're not in the prototype?
- 

For single states:

- _Single course_, _awaiting decision from provider_, interview scheduled_ need to be de-pluralised.
- _After interview, still awaiting decision_, take out "They will be in touch..." and "If you have been asked..."
- For _Course full, Provider did not respond in time_ or _offer withdrawn by provider_, can we be specific in the content to those scenarios? Not just cover all with "unsuccessful"? 
- _Accepted (pending conditions)_ and _Deferred_ are missing the indent for their guidance
- For Candidate did not respond in time, can we be specific and say: You didn't respond to your offers in time, so they have been automatically declined. "If now's the right time, you can still apply for courses that start this academic year.  To talk about your options, get in touch with a teacher training adviser."
- For _Candidate recruited_, should we end on a nice note "Congratulations on joining your teacher training course."
-

PR #455, opened 2021-02-17
Make the use of GOV.UK Notify optional

frankieroberto:
When switching to a new machine, it took me a good few minutes to decrypt this error message, caused by the lack of the Notify API key, and look up what the expected environment variable should be:

```bash
/apply-for-teacher-training-prototype/node_modules/notifications-node-client/client/api_client.js:20
    this.apiKeyId = arguments[0].substring(arguments[0].length - 36, arguments[0].length);
                                 ^

TypeError: Cannot read property 'substring' of undefined
```

This changes makes the use of Notify optional, enabling a faster start for designers and developers who don’t need to test email notifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ This prototype is based on the [GOV.UK prototype kit](https://github.com/alphago
 
 If you want to skip installing dependencies each time, you can instead type `npm run s`.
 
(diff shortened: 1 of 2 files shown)


paulrobertlloyd (approved):
❤️

PR #473, opened 2021-03-19
Fix all the things

paulrobertlloyd:
Went a bit crazy, tidied up code, fixed bugs, made things consistent, hopefully fixed reviewing application sections, made other qualifications section mandatory (even if you answer that you don’t want to add any other qualifications).

More to do around flight check and reviewing sections, but hopefully things are a bit more stable.


paulrobertlloyd:
@frankieroberto Don’t expect a full review, but feel free to kick the tyres.

PR #495, opened 2021-04-20
Update visa content

gregknight1:
Update of visa content after research.

Research suggested that the page was confusing - participants were not sure what visa 'sponsorship' is and were not are of the different types of providers - HEIs and SCITTs.

Changes:

- Introduce an H2 to break content up
- Explain what visa sponsorship is
- Explain the difference in provider types in terms of visa sponsorship
- Reduce the content about other visa routes, avoiding the long link text


frankieroberto:
Screen shot of the new content (to make it easier to review): 

apply-beta-p-residency--euz3j1 herokuapp com_application_12345_personal-information_residency_

app/views/application/personal-information/residency.njk
```diff
@@ -12,11 +12,14 @@
 {% block primary %}
 
 <p class="govuk-body">To get help with student visas and your immigration status, speak to a <a href="https://beta-adviser-getintoteaching.education.gov.uk/">Get Into Teaching</a> adviser.</p>
+<h2 class="govuk-heading-m">Getting a visa through your training provider</h2>
+<p class="govuk-body">You may be able to get a visa through your teacher training provider. This is called sponsoring a student visa.</p>
```
frankieroberto, line 16: Does "through" imply that the teacher training provider will get the visa for you? (Whereas I think you still have to do the application yourself?)

app/views/application/personal-information/residency.njk
```diff
@@ -12,11 +12,14 @@
...
+<p class="govuk-body">State funded schools, including School Centred Initial Teacher Training (SCITT) providers, cannot sponsor student visas for courses without a salary.</p>
+<p class="govuk-body">Contact your chosen training provider if you're not sure if they are university or a SCITT.</p>
 <p class="govuk-body"><a href="https://www.gov.uk/government/publications/register-of-licensed-sponsors-students">Check if a provider has a student visa sponsor licence</a>.</p>
-<p class="govuk-body">You may be eligible for another type of visa that does not need your training provider to act as sponsor.</p>
-<p class="govuk-body"><a href="https://beta-getintoteaching.education.gov.uk/international-candidates">Find out about student visas, visa routes that do not need a sponsor and visa options for trainees on courses with a salary</a>.</p>
+<p class="govuk-body">Find out more about your visa options on <a href="https://beta-getintoteaching.education.gov.uk/international-candidates">Get Into Teaching</a>.</p>
```
frankieroberto, line 22: Where we deep-link into Get Into Teaching, I think it’s a good idea to include some reference to the page we’re linking to within the link text, otherwise it can be a bit disorientating. (Better for screen readers too).

I’m also not sure that page is hugely helpful here, as it’s not _just_ about visas, and the visa section mostly just contains a link to GOV.UK anyway...?

frankieroberto (commented):
This seems like an improvement, but I wonder if this is now suggesting we need a bigger rethink? The "Getting a visa through your training provider" section doesn't seem to directly relate to helping people answer the "Do you already have the right to work or study in the UK?" question – as I understand it, it’s more content aimed at people who answer "Not yet, or not sure"?

I also wonder if the page is a little overwhelming with 3 links, and 4 calls to action?

gregknight1:
@frankieroberto Made the changes we discussed before around the Get Into teaching links. 

We also discussed that we would iterate this page after the work to structure visa sponsorship.

PR #503, opened 2021-05-14
Enable new link styles

paulrobertlloyd:
A new option in [GOV.UK Frontend v3.12.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0), that will be become the default in the next major version.

Do we want to enable these?


frankieroberto:
Enabled in #512 instead.

PR #511, opened 2021-06-24
Update course choice pages to match production

frankieroberto:
This applies a bunch of content and UI changes from the course choice flow in production to the prototype.

Changes:

* "Have you chosen a course to apply to?" changed to "Do you know which course you want to apply to?"
* "Pick a course location" changed to "Which location are you applying to?"
* Location names in radios lose their boldness
* Course selection changed from autocomplete to radios to match production
* Provider autocomplete gains a dropdown arrow
* Summary cards on review page are only 2/3rds width

I’ve quite possibly still missed some things. Also need to get my head around how the Apply -> Find -> Apply integration works.

### DIFF
--- a/app/routes/application/choices.js
+++ b/app/routes/application/choices.js
@@ -152,7 +152,7 @@ module.exports = router => {
         const l = {}
         l.text = loc.name
         l.hint = { text: loc.address }
-        l.label = { classes: 'govuk-label--s' }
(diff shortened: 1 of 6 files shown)


paulrobertlloyd:
Dirty work, but somebody’s gotta do it 😉 I haven’t looked to closely, but will assume it’s correct.

PR #515, opened 2021-06-25
Remove add another course interstitial

frankieroberto:
This removes the interstitial page after adding a course, which asks candidates if they would like to add another one:

Add-another-page

This page was added to encourage candidates to apply to multiple courses. However since then we have moved the "Add another course" secondary button to the top of the course choices page, and added some hint text to the "Have you completed this section?" radio button.

This also adds the same guidance text to just above the "Add another course" button:

course-choices

Our hypothesis is that this will serve the same purpose as the "Do you want to add another course" page, whilst removing an unnecessary extra page and giving users extra reassurance they just added the right course.

This change may also help to reassure candidates that they don’t need to add another course _right now_ but can instead come back to the Course choices page to add other courses later, before submitting the application.

An open question is whether we still need the "Success" banner or not – with an initial hypothesis being that we don't, as the page itself is a confirmation.

I’ve removed this line:

> After you submit your application you cannot apply for more courses through this service.

...as it's not actually true – you _do_ get a chance to apply again later through this service.

However this _might_ be the right page to explain the current business rules of only being able to apply again to 1 at a time within this cycle, only if they are all unsuccessful, or you withdraw them. But that's a complicated thing to explain!


paulrobertlloyd:
Makes sense to me. Half wondering if we can’t just remove the interstitial now on prod?

PR #519, opened 2021-07-07
Add end of cycle states

frankieroberto:
Add in end of cycle carry over messages:

### Applying again, but in a different cycle

Screenshot 2021-07-21 at 13 33 22

### Prompt to carry-over if there's an unsent application

Screenshot 2021-07-21 at 13 33 36


app/views/dashboard/_apply-for-next-year.njk
```diff
@@ -0,0 +1,27 @@
...
+    {% if numberOfOffersDeclined > 0 %}
+      You’ve declined {{ "your offer" if numberOfOffersDeclined == 1 else "all of your offers" }}.
+    {% elif numberOfApplicationsWithdrawn > 0 %}
+      You’ve withdrawn your {{ "application" if numberOfApplicationsWithdrawn == 1 else "applications" }}
+    {% else %}
+      Your {{ "application was" if (choices | length) == 1 else "applications were"}} unsuccessful
+    {% endif %}
+  </h2>
+
+  <p class="govuk-body">You submiited your application for courses starting in the 2021 to 2022 academic year, which have now closed.</p>
```
paulrobertlloyd, line 14:
```suggestion
  <p class="govuk-body">You submitted your application for courses starting in the 2021 to 2022 academic year, which have now closed.</p>
```

paulrobertlloyd:
When you cary over an unsubmitted application, we show a banner saying:

> We’ve copied your application. Please review all sections.

Is this necessary?

EmmaFrith:
> When you cary over an unsubmitted application, we show a banner saying:
> 
> > We’ve copied your application. Please review all sections.
> 
> Is this necessary?

oh do we? can we get rid? i don't think this should show

frankieroberto:
> > When you cary over an unsubmitted application, we show a banner saying:
> > > We’ve copied your application. Please review all sections.
> > 
> > 
> > Is this necessary?
> 
> oh do we? can we get rid? i don't think this should show. With both these carry over messages, the task list isn't mocked up as how it would look before apply reopening - I don't know how to do this.

Ah yeah, this wasn't intentional - the "copied" banner shouldn't be shown.

I've fixed this now, I think, so that both routes take you back to the application with no banner and in the "Find not yet open" state. If you change the `?findNotOpen=true` query to `findNotOpen=false` you can simulate Find being open but Apply being closed.

PR #522, opened 2021-07-14
Show visa sponsorship warning

frankieroberto:
For the next cycle, we will have course-level information about which courses candidates are able to get [visa sponsorship](https://bat-design-history.netlify.app/publish-teacher-training-courses/visa-sponsorship/) for.

Whilst hopefully candidates will spot this as part of browsing and filtering courses on Find, if they do add a course that does _not_ sponsor a visa AND they have said that they do not yet have the right to work or study in the UK, we can display a warning within their application journey.

This should hopefully avoid candidates applying to courses only to be rejected for visa reasons.

We would not block candidates from applying anyway though, as they may be able to obtain the right to work or study via some other route, or may already have this right, but be unsure or have answered the question wrongly.

To be able to display this warning as early as possible, the "Personal details" section has been moved to the top of the task list, as this is where the questions around nationality and the right to work and study are placed:

personal-details-first

The question around whether candidates have the right to work or study in the UK has been simplified, as the rest of the guidance will come later.

## Before

before

## After

after

Finally, if the candidate picks a course which doesn't sponsor a visa, and they don't have the right to work or study in the UK, they see this guidance within the Course summary box:

course-showing-warning


EmmaFrith:
Could we add a full stop?

Screenshot 2021-07-15 at 14 20 47

EmmaFrith:
Screenshot 2021-07-15 at 14 22 36

clutching at straws here, but maybe get rid of have to reduce word count?

EmmaFrith (approved):

PR #528, opened 2021-07-19
Update GCSE question and guidance

frankieroberto:
This aims to give candidates some guidance (or warnings) in the scenario that they:

* are currently (re)taking a GCSE, but are applying to a course which does not accept pending GCSE
* do not have a GCSE C or above and are not currently studying for one, and are applying for a course which does not accept equivalency tests

We also reviewed the content of the existing GCSE pages.

The initial GCSE question has been simplified, to remove reference to the C/4 grade, as this is caught later:

Screenshot 2021-07-23 at 09 21 00

If the candidates have a GCSE in English, the following page has been simplified, with the reference to English language as being preferred removed:

Screenshot 2021-07-26 at 15 44 14

If candidates select that they do not yet have this qualification, they get asked if they are currently studying for one:

Screenshot 2021-07-19 at 17 34 09

If they say yes, they can give details, as before:

Screenshot 2021-07-19 at 17 34 50

If they say no, they are show some guidance, and offered the chance to explain how they otherwise meet the standard:

Screenshot 2021-07-19 at 17 35 38

If they DO have a GCSE, but the grade is D / 3 or below, then they are asked this follow-up question:

Screenshot 2021-07-26 at 15 45 48

(If they say Yes, they can give details in a revealed textarea)

Depending on these answers, some guidance is shown in the Course Choice review pages:

Screenshot 2021-07-19 at 17 36 18

or

Screenshot 2021-07-19 at 17 37 10

### DIFF
--- a/app/data/session-data-defaults.js
+++ b/app/data/session-data-defaults.js
@@ -12,6 +12,19 @@ applicationWhereNotMeetingMinimiumDegreeRequirement.choices.FGHIJ.degreeRequired
 applicationWhereNotMeetingMinimiumDegreeRequirement.choices.ZYXWV.degreeRequired = "third"
 applicationWhereNotMeetingMinimiumDegreeRequirement.degree.abcde.grade = "Third-class honours"
 
+
+applicationWhereStudyingForGcse =  JSON.parse(JSON.stringify(require('./application')))
+
+applicationWhereStudyingForGcse.gcse.english.type = 'not-yet'
+applicationWhereStudyingForGcse.gcse.english.currentlyStudying = 'yes'
+applicationWhereStudyingForGcse.gcse.english.missing = 'I am currently studying for a GCSE English part-time.'
+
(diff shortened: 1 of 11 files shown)


frankieroberto:
One possible addition to this would be that IF you give your GCSE as a D/5 or below, then perhaps instead of just the free text box, we should first ask you the question of whether you're currently studying for a qualification (ie a resit / redo GCSE?).  That would then let us give the appropriate guidance in the course choice.

frankieroberto:
> One possible addition to this would be that IF you give your GCSE as a D/5 or below, then perhaps instead of just the free text box, we should first ask you the question of whether you're currently studying for a qualification (ie a resit / redo GCSE?). That would then let us give the appropriate guidance in the course choice.

I've done this now, with the question currently being "Are you currently studying to re-take your GCSE in [subject]?"

PR #529, opened 2021-07-21
GCSE questions and guidance – further iteration

frankieroberto:
This builds on #528 but also splits the first question into two, to simplify it for the majority of people who _do_ have a GCSE.

(Separate branch for now in case we decide against this)

## First question

> Screenshot 2021-07-21 at 15 15 10

## If they do not have a GCSE

> Screenshot 2021-07-21 at 15 16 26


frankieroberto:
Going to close this, as we’ve decided to try and stick to having a single initial question for now.

PR #530, opened 2021-07-27
Half-update references section, minus selection flow

paulrobertlloyd:
We’ve changed a bit of the content around references, as well as the status indicators. But we haven’t prototyped the new reference selection process.

This PR updates the application menu and other pages to bring them closer inline with what’s on (or soon will be on) prod, but without adding the select 2 references flow.

Is that a problem do we think?


frankieroberto (approved):
👍  thanks for (half-) updating the prototype.

We can add the "Select 2 references" section at some point in the future, maybe.

PR #531, opened 2021-07-28
Add questions to determine fee status

paulrobertlloyd:


frankieroberto:
Quick comments from me:

* Hard to interpret "What is your permanent address?" - are any addresses permanent? 😉 
* Do we need a full address for "permanent address", or just a country?
* 6 input fields possibly too many to be conditionally revealed? ([guidance](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question) now says "if the related question is complicated or has more than one part, show it on the next page in the process instead")
* Possibly change "In the UK" / "Outside the UK" to "UK" / "Another country" to be more inclusive?
* The "Change" link for contact address links to country question page, which is possibly confusing?

PR #539, opened 2021-08-18
International candidate journey improvements

paulrobertlloyd:
- [x] Capture if candidate has EU (pre-)settled status
- [ ] Capture date of entry into the UK


app/routes.js
```diff
@@ -37,6 +37,9 @@ require('./routes/survey')(router)
 // Clear all data in session if you open /admin/clear-data
 router.post('/admin/clear-data', function (req, res) {
   req.session.data = {}
+
+  console.log('hello')
```
frankieroberto, line 41: 👋

paulrobertlloyd: Oops. Was testing why the clear data page no longer works. It’s got me flummoxed!

B-Bonner:
**Dual nationality**

If you select British and another nationality, or Irish and another nationality, it should ask you if you’ve lived in the UK for the last 3 years. At the moment it takes us straight to the summary page.

B-Bonner:
Maybe we should change the last summary line to 'Will you have lived in the UK for 3 years or more?' or 'Lived in UK for 3 years of more before course starts', to make it match the question.

Screenshot 2021-08-24 at 15 55 49

B-Bonner:
1) The immigration status should disappear if you say you 'don't have the right yet' (maybe that's a prototype bug?)

2) Can we replace 'gain' with 'get' in the last summary line? To make it match the question.

3) The immigration status didn't update when I changed the nationality to French and the status to 'pre-settled' (again, perhaps a bug?)

Screenshot 2021-08-24 at 16 42 18

B-Bonner:
Should this question be something like 'Will you have lived in the UK in the 3 years leading up to your teacher training?' or 'Will you have lived in the UK in the 3 years before your teacher training starts?'

Because we're talking about the years immediately before the training, right? If I'm British and lived in the UK from birth, but decided to live in Spain for the last 2 years, I would have 'lived in the UK for 3 or more years', but wouldn't be eligible for home fees.

The hint text could say something like: 'This includes any time spent temporarily in the EU, EEA or Gibraltar.' I think we need to get more details about this from the international team today, because we don't know how long constitutes as "temporary" or what's considered a "valid reason".

Screenshot 2021-08-24 at 15 40 22

PR #540, opened 2021-08-26
Drop text advising candidates speak to teacher training advisers about immigration

frankieroberto:
Teacher training advisers cannot usually help with this, as all they do is point candidates towards the Home Office.

### DIFF
--- a/app/views/application/personal-information/immigration.njk
+++ b/app/views/application/personal-information/immigration.njk
@@ -11,9 +11,6 @@
 
 {% block primary %}
   {{ govukRadios({
-    hint: {
-      html: "To get help with student visas and your immigration status, speak to a <a href=\"https://beta-adviser-getintoteaching.education.gov.uk/\">teacher training adviser</a>."
-    },
     items: [{
       value: "Yes",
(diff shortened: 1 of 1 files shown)


B-Bonner:
I think we either need to reinstate the 'not sure' option we had before ('Not yet, or not sure'), or we need to somehow link to more information. 

Because, although you'd know if you have pre-settled or settled status, you may not know that you **don't** have the right to work/study in the UK if you haven't looked into it yet.

frankieroberto:
@B-Bonner 

> I think we either need to reinstate the 'not sure' option we had before ('Not yet, or not sure'), or we need to somehow link to more information.
> 
> Because, although you'd know if you have pre-settled or settled status, you may not know that you **don't** have the right to work/study in the UK if you haven't looked into it yet.

The difficulty with doing this is that if you answer "Not yet", we now ask a follow-up question about how the candidates will get the right to work and live in the UK (so that we can advise them if they need a sponsored visa and apply for a course which can't sponsor visas).

We could add a "Not sure" option, but then we'd have to not show that guidance? 🤔

PR #548, opened 2021-09-14
Make it clear you can return to reference

frankieroberto:
We currently tell referees that they need to complete the reference in a "one sitting". However they _can_ save and come back to it at any time, using the original link sent to their email.

Feedback from referees has indicated that they find the requirement to complete the reference in a single sitting to be stressful, especially given that the reference takes longer than they initially expect.  Some referees may work around this by copying and pasting to and from a separate file, but this can be cumbersome, and means they might not see the word count straight away.

To address this, we are changing the content to make clear that they can return to the form after saving it. The buttons are also relabelled from "Continue" to "Save and continue". However the button on the page with the 500 word textarea is just labelled "Save" so that it's clear you can save this even if you aren't ready to continue.

The check your answers page also has new content to make it clearer that referees can exit at this page and return to continue, rather than submit it now, and the title changed from "Check your answers before submitting your reference" to "Your reference for [candidate name]".

The title of the reference page has changed from "Your reference for [candidate name]" to a question: "Does Jane Doe have the potential to teach?". This is to distinguish it from the last page, and because elsewhere we use "reference" to refer to the whole journey, not just this question.

The bullet points on the reference page have also been revised to be more succinct, and more in keeping with the style guide. (Previously the bullets contained headings and paragraphs).

We should monitor this to see if it has any impact on the average time it takes referees to complete a reference, as one possible negative side-effect is referees taking longer to submit instead of rushing a reference because of the perceived "single sitting" requirement.

We are also separately changing the content of the referee email: https://github.com/DFE-Digital/apply-for-teacher-training/pull/5582

## Before

Screenshot 2021-09-15 at 15 12 18

Screenshot 2021-09-14 at 14 39 01

Screenshot 2021-09-15 at 14 15 09

## After

Screenshot 2021-09-15 at 15 12 01

Screenshot 2021-09-14 at 14 38 53

Screenshot 2021-09-15 at 14 39 43

### DIFF
--- a/app/views/admin/index.njk
+++ b/app/views/admin/index.njk
@@ -15,5 +15,6 @@
     <li><a href="/admin/flags">View feature flags</a></li>
     <li><a href="/admin/send-email">Send email</a></li>
     <li><a href="/admin/states">Mock application states</a></li>
+    <li><a href="/reference">Giving a reference flow</a></li>
   </ul>
 {% endblock %}
--- a/app/views/reference/comments.njk
+++ b/app/views/reference/comments.njk
@@ -2,7 +2,7 @@
 
 {% set formaction = referrer or "/reference/review" %}
 {% set candidate_name = "Jane Doe" %}
-{% set title = "Your reference for " + candidate_name %}
+{% set title = "Does " + candidate_name + " have the potential to teach?" %}
 {% set hasAccountLinks = false %}
 
 {% block pageNavigation %}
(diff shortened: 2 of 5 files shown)


frankieroberto:
[Referees need to know they can save and come back later](https://trello.com/c/0T75uQdC/3945-referees-need-to-know-they-can-save-and-come-back-later)

EmmaFrith:
what happens when they click 'save an continue' @frankieroberto?

frankieroberto:
> what happens when they click 'save and continue' @frankieroberto?

No change - ie they go to the next question, and then to the review page. But at least it's clearer that it has been saved, and the can return to question via back link or review page?

Perhaps a future enhancement could be some kind of "Save" OR "Save and continue" feature. Or perhaps we autosave every few seconds in the background (either to their local machine, or remotely).  But I figured this would at least help to make clear that they don't have to write and submit the reference in a "single sitting"?

EmmaFrith:
what IS the next question? sorry I have no idea how to access the form! @frankieroberto

frankieroberto:
> what IS the next question? sorry I have no idea how to access the form! @frankieroberto

Ah, sorry! I only just discovered that the reference journey is also in the prototype...

Here's the existing flow: https://apply-beta-prototype.herokuapp.com/reference
Here's the proposed changes: https://apply-beta-p-make-it-cl-ltrdeo.herokuapp.com/reference

EmmaFrith (approved):

PR #575, opened 2021-10-15
Tweak to decline reference flow

frankieroberto:
We have some evidence (from support) of referees declining references by mistake.

We think this is because the email contains two links, and they click on the second one (to decline), and then select "Yes, I’m sure" without reading the rest of the page (or skim-reading it or mis-interpreting it).

We hope to avoid this by changing the wording of the radio buttons so that:

* the "Yes" action corresponds to giving a reference instead of declining
* the radio button labels contain a full answer, so that they can be understood without reading the question above

## Before

Screenshot 2021-10-15 at 12 41 15

## After

Screenshot 2021-10-18 at 11 27 09

### DIFF
--- a/app/views/reference/decline.njk
+++ b/app/views/reference/decline.njk
@@ -6,8 +6,8 @@
 {% set hasAccountLinks = false %}
 
 {% block primary %}
-  <p class="govuk-body">Teacher training courses can fill up quickly, and candidates need 2 references to apply.</p>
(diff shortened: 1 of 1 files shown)


EmmaFrith:
i'm also thinking we could add a free text box if someone selects the decline option - 1. to reduce the possibility of someone doing it accidentally 2. so we learn why people decline

PR #580, opened 2021-10-26
copy tweaks

EmmaFrith:
A few copy changes following initial peer review


frankieroberto:
I've moved the sentence out from the hint text, back to the bottom. We don't think it's necessary for this to be read before answering the question.

Screenshot 2021-11-01 at 17 05 21

Have also set a threshold on the word count, so that you don't see it until hitting 85% of the limit.

PR #591, opened 2021-11-04
Simplify "other qualifications" flow

frankieroberto:
This simplifies the "A levels and other qualifications" flow so that the "What type of qualification do you want to add?" no longer contains an "I do not want to add any A levels and other qualifications" answer, in order to reduce the complexity of the question, and to more easily represent the candidate’s answers on the application review and within Manage.

Instead, candidates are routed through the review page, which contains the guidance about why it is a good idea to add A levels and other qualifications. This also means that this content is still visible even after adding the first qualification, to remind candidates that they can continue to add other qualifications.

Additionally, on the details screen, the font size of the labels has been reduced, and the "Do you want to add another qualification?" question has been dropped (although this could be restored).

➡️ [Trello card](https://trello.com/c/ugz3MWoX/4095-help-candidates-enter-other-qualifications-or-not-more-easily)

### Before

before

before-2

### After

screenshot-localhost_3001-2021 11 15-10_44_00

screenshot-localhost_3001-2021 11 15-10_44_30

after-3

screenshot-localhost_3001-2021 11 15-10_57_34

### DIFF
--- a/app/routes/application/other-qualifications.js
+++ b/app/routes/application/other-qualifications.js
@@ -67,16 +67,6 @@ module.exports = router => {
     const { year, org, type } = application.otherQualifications[id]
     const { typeUk } = application.otherQualifications[id]
 
-    if (next === 'add-type') {
-      application.otherQualifications[nextId] = {
-        year,
-        org,
-        type,
-        typeUk
-      }
-      res.redirect(referrer || `/application/${applicationId}/other-qualifications/${nextId}/details`)
-    } else {
-      res.redirect(referrer || `/application/${applicationId}/other-qualifications/${next}`)
-    }
+    res.redirect(referrer || `/application/${applicationId}/other-qualifications/${next}`)
   })
 }
--- a/app/routes/application/other-qualifications.js
+++ b/app/routes/application/other-qualifications.js
@@ -4,44 +4,50 @@ module.exports = router => {
   // Generate new id and redirect to that qualification
   router.get('/application/:applicationId/other-qualifications/add', (req, res) => {
     const id = utils.generateRandomString()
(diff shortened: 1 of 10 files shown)


adamsilver:
I would avoid the complexity of multiple controls/buttons fighting for attention.

Screen one before the candidate has interacted.

```
Heading: A levels and other qualifications</h1>

<Prose go here>

Do you want to add A levels or other qualifications?
( ) Yes
( ) No

[Continue]
```

Once they answer this question/complete the flow they can then mark the section complete or not.

frankieroberto:
> I would avoid the complexity of multiple controls/buttons fighting for attention.

I had a similar thought - this is also what we do on the Work history flow where there's an initial guard question.  We'd still have the 2 buttons once you're on the review page though, as the secondary button would be used to add additional qualifications, and we'd possibly also still show the same guidance, so I was tempted to skip the guard question...

The guard question also introduces the complexity that if they answer "No", we'd have to then show a review page summarising the single question the've just answered, and asking them to mark the section as complete.  But maybe that's ok (and we do the same on the Interview needs section)... 🤔

frankieroberto:
@adamsilver I've updated this to add an initial filter question now (see updated screenshot, or linked review app). Think this works well... If you delete all your qualifications after initially adding them, you're redirected back to the filter question (rather than presenting the review page with nothing on it).

EmmaFrith:
Have added a commit with some proposed copy tweaks - screenshots below:

Screenshot 2021-11-12 at 12 24 01
Screenshot 2021-11-12 at 12 24 25
Screenshot 2021-11-12 at 12 25 18
Screenshot 2021-11-12 at 12 26 21

Let me know what you think @frankieroberto

PR #593, opened 2021-11-10
Update immigration questions following review

frankieroberto:
Changes made:

* Incorporate radio hint text into the label
* Change "Other" to "Another..."
* Change "How will you get the right to work or study in the UK?" to "Will you need the course provider to sponsor a skilled worker or student visa?"
* Change immigration status example from "permanent residence" to "Indefinite leave to remain"
* Change "Give any relevant information" to "Immigration status"
* Add threshold to textareas

# Screenshots

| Before | After |
| ---- | ----- |
|  before-visa-route | Screenshot 2021-11-10 at 15 48 16 |
| before-immigration-status |  after-immigration-status |


adamsilver (dismissed):
I really like how you changed this into a yes no answer...

Some issues/questions

(1) Will all candidates understand that they will “need a course provider to sponsor a skill worker or student visa”?

(2) Related to (1) I think we could do with adding "I don't know" as an answer

(3) Instead of "give details" consider "Tell us how you intend to get the right to work or study" (or similar)

(4) Related to (3) you can just make the label ‘No’

(5) Change “What is your...” to “What’s your...”

(6) Add a "not sure" to the "What is your immigration status" question

(7) Is the "immigration status" textarea question optional?

frankieroberto:
> (1) Will all candidates understand that they will “need a course provider to sponsor a skill worker or student visa”?

Hopefully! We added some guidance on this to both Find and on Get into teaching. However we don't know for sure. We did consider linking to the guidance, or repeating some of it. But let's see if that's needed or not first?

> (2) Related to (1) I think we could do with adding "I don't know" as an answer

The main aim of this question is to avoid the preventable rejection of people applying to courses which cannot sponsor visas, when they need one.  Adding an "I don't know" option makes this more difficult (as we can’t then display such a strong warning), and might mean that providers reject them on this basis anyway.

Initial research seems to suggest that international students likely have a good understanding of what their visa route options are. However if we get evidence that it’s still confusing, we could either add a third option, or give candidates some additional guidance so that they know how to answer.

> (3) Instead of "give details" consider "Tell us how you intend to get the right to work or study" (or similar)

Yeah, I wasn't sure about this either. Will have a think.

> (4) Related to (3) you can just make the label ‘No’

Will have a think on this. Think it's good to give the Family visa and Graduate visa as examples, as they may not realise that Graduate visas do not need to be sponsored by the providers.

> (5) Change “What is your...” to “What’s your...”

👍 

> (6) Add a "not sure" to the "What is your immigration status" question

If someone isn't sure, or it's complicated, they could write it into the free text box?  Probably better that we let them explain it in their own words rather than just submitting a "Not sure" to the provider?

> (7) Is the "immigration status" textarea question optional?

No.

EmmaFrith (approved):

PR #599, opened 2021-11-25
Prevent application submission where only Foundation level degrees have been entered

frankieroberto:
This updates the prototype to prevent candidates from marking the degree section as complete if they've only added Foundation degrees.

To do this, the "Have you completed this section?" question is removed from the bottom, and the "Add another degree" buttons switches to the primary green colour (rather than secondary grey). The guidance is also shown.

## Screenshot

screenshot-localhost_3000-2021 11 29-15_37_47

### DIFF
--- a/app/data/session-data-defaults.js
+++ b/app/data/session-data-defaults.js
@@ -24,6 +24,24 @@ const applicationWithDegreeSectionNotStarted = JSON.parse(JSON.stringify(require
 applicationWithDegreeSectionNotStarted.completed.degree = null
 applicationWithDegreeSectionNotStarted.degree = {}
 
+const applicationWithFoundationDegreeOnly = JSON.parse(JSON.stringify(require('./application')))
+applicationWithFoundationDegreeOnly.completed.degree = null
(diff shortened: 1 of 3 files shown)


EmmaFrith:
could we use the same term here and here? so that the user can automatically see which thing they're looking for based on the guidance they've just seen
Screenshot 2021-11-29 at 13 31 24
Screenshot 2021-11-29 at 13 32 27

EmmaFrith:
Should we say something here that makes it obvious that it's ok if you're still studying for your degree? e.g. you can still apply if you haven't got your grade yet 

Screenshot 2021-11-29 at 13 34 46

frankieroberto:
Thanks @EmmaFrith!

We’ve review the guidance and updated it to:

> You need a bachelor’s degree or equivalent to start teacher training.
> Add your bachelor’s degree even if you have not got your grade yet.
> You can also add any other degrees that you have, or you’re working towards.

...having decided that it was better to consistently use "bachelor" than "undergraduate" (which is a more loosely-defined term which technically might include a Foundation course).

PR #603, opened 2021-12-03
Add fee information to offer details

frankieroberto:
This would initially just copy across the content from the Fees or Salary section on Find.  The amount of content can vary quite a bit between providers.

This information is required as part of the pre-contract information by the regulations. We probably fulfil that already through the link to the course page on Find, plus any additional information given direct to candidates by providers off-service. 

However, it still feels like a good idea to repeat this on the offer page, to remind candidates of the fees / salary at the point of accepting an offer.

## Fees

Screenshot 2022-06-23 at 10 36 40

## Salary

Screenshot 2022-06-23 at 10 36 53


EmmaFrith:
Would we need to structure the fees section on Find more to be able to do this effectively? 

For example, I can see that not all course pages show the international candidate fee: 

Screenshot 2022-01-04 at 14 27 20

EmmaFrith:
A less sophisticated, but more MVP solution, could be to just add a line here to be explicit about checking course fees. - e.g.:

'Check you're happy with the [course fees](link to fee section on relevant course page) before accepting the offer.'

Screenshot 2022-01-04 at 14 35 39

EmmaFrith:
Maybe out of scope for this PR, but we might also want to think about whether these offer details should be included in the summary card on the dashboard, rather than hidden on a separate page. 

At the moment, the link text 'Respond to offer' doesn't make it sound like you'll get to see all the offer details. 

We could also just change the link text so it's clearer that you're reviewing the offer details before responding. 

Screenshot 2022-01-04 at 14 54 59

EmmaFrith:
Just an FYI, I'm working on a related ticket: https://trello.com/c/NYXJ8r9p/4258-dev-edit-content-on-offer-details-page

frankieroberto:
> A less sophisticated, but more MVP, solution could be to just add a line here (see green arrow) to be explicit about checking course fees. - e.g.:
> 
> 'Check you're happy with the [course fees](link to fee section on relevant course page) before accepting the offer.'

I don't think we can add a link within the radios (for accessibility reasons), but we could add a generic line elsewhere within the page, instead of copying across the content from Find. The line would either have to cover both salaried and fee-paying courses, or have some logic to switch between the 2 versions. There is already a link to the course page (from the course title), so I think it’d be better to make this more explicit somehow, rather than duplicate it?

That said, I think it's also worth continuing to explore pulling through all the fee/salary information from Find, and then we can evaluate both options?

Also agree that longer-term we should look at structuring the fee/salary information a bit more.

PR #604, opened 2021-12-03
Add note on full terms and conditions, and being able to withdraw

frankieroberto:
This explores how we might make it clearer to candidates that accepting an offer means accepting the full terms and conditions from the provider (which they should have received off-service) and that they can cancel/withdraw after they accept (which they have the right to do under consumer regulation).

Feedback on the content welcome.

## Before

Screenshot 2021-12-03 at 15 36 27

## After

Screenshot 2021-12-03 at 15 35 43

### DIFF
--- a/app/views/application/decision/view.html
+++ b/app/views/application/decision/view.html
@@ -18,7 +18,7 @@
     {% if numberOfOffersReceived > 1 %}
       <li>your other {{ "offer" if numberOfOffersReceived == 2 else "offers"}} will be automatically declined</li>
     {% endif %}
-    <li>you will be able to cancel and withdraw at any time up until the course has started</li>
+    <li>you will be able to withdraw at any time up until the course has started</li>
   </ul>
 
   <p class="govuk-body">If you decline all of your offers, or do not respond in time, you can still apply for courses that start this academic year.</p>
--- a/app/views/application/decision/view.html
+++ b/app/views/application/decision/view.html
@@ -12,14 +12,17 @@
 {% block primary %}
   {% include "_includes/item/offer.html" %}
 
-  <p class="govuk-body">Before you accept:</p>
+  <p class="govuk-body">If you accept this offer:</p>
 
   <ul class="govuk-list govuk-list--bullet">
     {% if numberOfOffersReceived > 1 %}
-      <li>if you accept this offer, your other {{ "offer" if numberOfOffersReceived == 2 else "offers"}} will be automatically declined</li>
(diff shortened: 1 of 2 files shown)


EmmaFrith:
A few questions:

1. What do we mean by 'cancel and withdraw'? Can we just say 'withdraw' to simplify it? 
2. Is it definitely/always true that 'you can cancel and withdraw at any time up until the course starts'?
3. Do we know that providers do provide terms and conditions? Does DfE require providers to do this?

EmmaFrith:
I'm also working on a related PR because I think this whole bit of content is overly complicated. See: https://github.com/DFE-Digital/apply-for-teacher-training/pull/6289

Maybe we need to think about this whole page in the round.

frankieroberto:
> 1. What do we mean by 'cancel and withdraw'? Can we just say 'withdraw' to simplify it?

Good point, let's simplify. I did wonder whether "withdraw" might feel odd after accepting an offer but before starting the course, but it's what we use in the interface elsewhere.

> 2. Is it definitely/always true that 'you can cancel and withdraw at any time up until the course starts'?

I'm not 100% sure, no. We'll have to clarify this somehow. As I understand it, there's a 14 day cancellation period under some regulations, but there might also be separate rights to cancel up until enrolment? Also it may vary by provider, fee paying vs salaried, or other factors?

> 3. Do we know that providers do provide terms and conditions? Does DfE require providers to do this?

I don’t think it’s required by DfE, but it seems to be required by the Competition and Markets Authority guidance, at least for higher education providers.

app/views/application/decision/view.html
```diff
@@ -12,15 +12,18 @@
...
-  <p class="govuk-body">Before you accept:</p>
+  <p class="govuk-body">If you accept this offer:</p>
 
   <ul class="govuk-list govuk-list--bullet">
     {% if numberOfOffersReceived > 1 %}
-      <li>if you accept this offer, your other {{ "offer" if numberOfOffersReceived == 2 else "offers"}} will be automatically declined</li>
+      <li>your other {{ "offer" if numberOfOffersReceived == 2 else "offers"}} will be automatically declined</li>
     {% endif %}
-    <li>if you decline all of your offers, or do not respond in time, you can still apply for courses that start this academic year</li>
+    <li>you will be able to cancel and withdraw at any time up until the course has started</li>
```
frankieroberto, line 21:
```suggestion
    <li>you will be able to withdraw at any time up until the course has started</li>
```

PR #605, opened 2021-12-03
Link to terms and conditions

frankieroberto:
This builds about #604 but adds a link to the full terms and conditions from the provider.  To do this, we’d need to collect those conditions from providers (as a link or uploaded document) via Publish, so this would be a longer-term project.

## Screenshot

Screenshot 2021-12-03 at 15 42 46


JonF1 (dismissed):
.

PR #640, opened 2022-02-18
Update personal statement feedback in apply again

frankieroberto:
The way that feedback is being collected when rejecting a candidate is changing.

Previously there were separate "Personal statement" and "Subject knowledge" sections beneath a "Quality of application" heading. When these were filled in, we marked the "Why you want to teach" and the "Your suitability to teach a subject or age group" sections as needing a review, and displayed the relevant feedback within those sections.

The new feedback form has fields for "Quality of writing" and "Other" under a "Personal statement" heading.

If either of these two fields are filled in, we would now show both the "Why you want to teach" and the "Your suitability to teach a subject or age group" sections as needing review, and show the same feedback in both sections.

The headings used have also changed to match the new feedback interface.

### Feedback on dashboard

before-1

### Feedback on personal statement section within apply again

before-2

### Feedback on dashboard

after-1

### Feedback on personal statement section within apply again

after-2


adamsilver:
Not part of this PR directly but I the link text would be better as `get in touch with a teacher training advisor` as opposed to `teacher training advisor`.

I'm wondering if it would be tidier and more consistent to lose the quotes and do something more like this:

```
## Feedback from previous applications

### Gorse Scitt

> Quality of writing:
> 
> Blah blah blah
> 
> Other:
> 
> Blah blah blah

### Other training provider

> etc etc
```

EmmaFrith:
Thanks @frankieroberto this looks good. I agree that we probs don't need both quotes and inset text. Just thinking that we're using inset text A LOT - how about just reserving it for the actual quotes. E.g. 

#Feedback from previous applications

##Sheffield University 

Quality of application:

^ You were bad. 

Other 

^ You were bad. 

##Nottingham Uni

Quality of application:

^ Good effort.

frankieroberto:
@EmmaFrith @adamsilver thanks both. I do like the idea of using inset text to denote quotations /  another 'voice', rather than using speech marks. Without the inset text across the whole section, we might need something else though to separate it out from the rest of the page. Especially as otherwise you then get 3 heading levels next to each other.

You both differed in whether you thought the "Quality of writing:" and "Other:" lead-ins should be part of the inset text or not. I've mocked up both below.  I still think that we can probably lose them altogether in this context, and just include the content, as "Other" isn't really adding and value, and I'd expect that "Quality of writing" is explained within the actual text anyway too.

## Option 1
(Ignore the long line lengths here, I forgot to wrap them)

option-1

## Option 2

option-2

frankieroberto:
> Yea there's a lot of headings going on. This in particular grosses me out

Agreed!

@EmmaFrith and I did a bit of pair design and have refined it to this, which drops the "Feedback from previous applications" heading and also the "Quality of writing" and "Other" lead-in lines:

Screenshot 2022-02-21 at 11 18 12

EmmaFrith (approved):

PR #648, opened 2022-04-21
Prototype kit update

frankieroberto:
Updates the GOV.UK Prototype Kit from version 11 to version 12.

The main things this changes are:

* adds a new password screen, replacing basic authentication (see [12.0.3 update](https://github.com/alphagov/govuk-prototype-kit/releases/tag/v12.0.3))
* switches to dart sass (from node-sass) for better compatibility with M1 Macs (see [12.0.2 update](https://github.com/alphagov/govuk-prototype-kit/releases/tag/v12.0.2))
* updates to version 4.0.0 of govuk-frontent (see p12.0.0 update](https://github.com/alphagov/govuk-prototype-kit/releases/tag/v12.0.0))

➡️  See [Preview of the changes](https://apply-prototype-kit-upgrade.herokuapp.com)


frankieroberto:
Some notes from the upgrade:

I upgrade using the [instructions on the GOV.UK Prototype Kit website](https://govuk-prototype-kit.herokuapp.com/docs/updating-the-kit) which were super helpful and I appreciated being able to go through it step by step, and to have the script to run.

The only issues I experienced were:

1. The script showed the message: "ERROR you must run update.sh in a folder containing a GOV.UK Prototype Kit installation", even though I was in the folder containing the prototype. I downloaded the script to go through it line by line, and realised that it [was checking for the phrase `govuk-prototype-kit` in the `package.json` file](https://github.com/alphagov/govuk-prototype-kit/blob/main/update.sh#L33). This wasn't present in ours as we had renamed it to be specific to the app. I could workaround this by temporarily renaming it back to `govuk-prototype-kit`.

2. When running `npm start` I got a few missing module errors, but these were easy to fix by adding them back in, as the update instructions suggest.

3. The password screen didn't initially work as we had renamed our `layout.html` file to `_layout.html`. I removed the underscore prefix and updated all the references to it in our views.

4. I also had to copy `unbranded.scss` back into our repo, as we'd previously deleted it.

5. I had to make some manual changes to the `server.js` file to add back in some customisations that we'd made previously. These add references to `app/utils/filters-with-app-context` which adds some filters that have access to the `app` context, and some lines which set some data to `app.locals`.

Other than that, it worked smoothly and took less than half an hour. 👍

PR #709, opened 2022-09-23
Draft: update disability and health conditions

frankieroberto:
This is a draft update of the disability questions asked when submitting an application form.

Currently, candidate are first asked "Are you disabled?". If they answer "Yes", they are asked "What disabilities do you have?" with a set of checkboxes.

The [updated HESA guidance](https://www.hesa.ac.uk/collection/c22053/a/disable) includes a suggestion from Advance HE which includes a detailed explanation of the definition of disability under the Equality Act 2010, and then asks:

> Considering the above, do you have an impairment, health condition or learning difference that has substantial and long-term impact on your ability to carry out normal day-to-day activities?

In our own research, one participant with dyslexia also mentioned that not everyone with dyslexia considers themselves to be disables (although they answered Yes to the question).

Rather than update the initial Yes / No question wording, one approach might be to skip straight to the checkboxes question, and add a "none of the above" type answer.


frankieroberto:
Work in progress screenshot:

disability-wip

frankieroberto:
Merged into #714

PR #726, opened 2022-10-11
Show message if you can't apply to a course

frankieroberto:
This would allow candidates to add courses which aren't yet open for applications, but would warn them and prevent the application being submitted.

This would hopefully make it more obvious (than just not allowing them to find the course in the dropdowns/radios), and allows candidates to get their whole application ready, waiting for the course to open. Alternatively they can choose a different course.

We can use the same approach for when a course is "full" / has no vacancies.

🗂️ [Trello card](https://trello.com/c/l7u2mGiK/762-allow-candidates-to-add-courses-not-yet-open-to-their-application-but-not-submit-the-application-yet)
➡️  [Preview](https://apply-tt-pr-726.herokuapp.com/application/19415)

## Course choice page / application review

courses page

## Content replacing the submit button on the application review page

Screenshot 2022-10-11 at 15 02 29

### DIFF
--- a/app/views/application/review.html
+++ b/app/views/application/review.html
@@ -145,9 +145,9 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       {% if cannotSubmitReasons.length == 1 %}
-        <p class="govuk-body">You cannot submit this application as {{ cannotSubmitReasons[0] }}.</p>
+        <p class="govuk-body">You cannot submit this application because {{ cannotSubmitReasons[0] }}.</p>
       {% else %}
-        <p class="govuk-body">You cannot submit this application as:</p>
+        <p class="govuk-body">You cannot submit this application because:</p>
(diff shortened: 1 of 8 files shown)


app/views/application/review.html
```diff
@@ -116,7 +116,45 @@ <h3 class="govuk-heading-m">References to be requested if you accept an offer</h
...
+  {% endfor %}
+
+  {% if canSubmit %}
+    {{ govukButton({
+      text: "Continue"
+    }) if not closed }}
+  {% else %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      {% if cannotSubmitReasons.length == 1 %}
```
frankieroberto, line 147: Think this needs to use the Nunjucks filter.
```suggestion
      {% if cannotSubmitReasons | length == 1 %}
```

PR #727, opened 2022-10-11
Survey users on whether provider rejection feedback is helpful

frankieroberto:
This would add some buttons to allow candidates to say whether the feedback given when they are rejected is helpful or not.

🗂️ [Trello card](https://trello.com/c/Mg1wH7xS/700-ask-candidates-whether-providers-rejection-feedback-was-useful)

In the dashboard, a single question would be shown beneath the feedback:

Screenshot 2022-10-11 at 16 45 57

Clicking a button would replace the buttons and text with a confirmation message:

Screenshot 2022-10-11 at 16 46 35

...or if they said it was not helpful:

Screenshot 2022-10-11 at 16 47 02

(Ideally we'd do this via javascript to avoid a page reload)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,21 +7,19 @@ document.addEventListener('DOMContentLoaded', function () {
   window.GOVUKFrontend.initAll()
 })
 
-
-function setupFeedbackButtons() {
+function setupFeedbackButtons () {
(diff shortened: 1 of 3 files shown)


frankieroberto:
[Ask candidates whether providers' rejection feedback was useful](https://trello.com/c/Mg1wH7xS/700-ask-candidates-whether-providers-rejection-feedback-was-useful)

PR #746, opened 2022-12-09
Boiler plate content for reasons for rejection

MaeveRoseveare:
When a provider rejects a candidate, we want to show some text to help the candidate take the next step and keep them in the application journey. We also know that sometimes the rejection reasons from providers is not helpful.
Trello ticket: https://trello.com/c/zmcV9v7v/495-r4r-draft-boiler-plate-feedback-text

When all 3 applications are unsuccessful, we would show some 'boiler plate' content above the applications to help them onto the next step.

Image below shows the text above the rejected applications, each bullet point would only show depending in what 'rejection reason' a provider chooses. From analysis, the majority of providers only ever select 1 or 2 reasons. So in the screenshot below a candidate would mostly only see 1 or 2 bullet points (I've included more just to see what it would look like).

Screenshot 2022-12-09 at 14 34 08

### DIFF
--- a/app/routes/application/dashboard.js
+++ b/app/routes/application/dashboard.js
@@ -334,23 +334,21 @@ module.exports = router => {
       case 'ended-without-success':
         choices.ABCDE.status = 'Unsuccessful'
         choices.ABCDE.feedback = {
-          personalStatement: {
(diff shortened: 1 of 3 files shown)


MaeveRoseveare:
[R4R: Draft boiler plate feedback text](https://trello.com/c/zmcV9v7v/495-r4r-draft-boiler-plate-feedback-text)

frankieroberto (approved):

PR #757, opened 2023-01-13
Adding SKE info to other places in apply

MaeveRoseveare:
Exploring where else we could add info about subject knowledge enhancement (SKE) courses.

We've looked at various places we could add this information, and mocked up one option on the 'Suitability to teach a subject or age group' - this feels like the only place we could add more info. See screenshot below. 

However, it still doesn't feel quite right to say this here, it has nothing to do with what we're asking candidates to do here. Also, not everyone can do an SKE, so do we want to send people away at this point to the GIT page - rather let them fill out the form.

If we did conditional content, it feels like a lot of work for a very small amount of text, that we're not sure people will even read. We could only do conditional text if we know what course the candidate has chosen, however there is no guarantee that a candidate might start this section before choosing a course.

Screenshot 2023-01-13 at 13 58 15

Further comments about the other areas we explored are in this trello ticket: https://trello.com/c/rc1jMLja/1075-investigate-where-else-we-could-surface-ske-information-for-candidates

### DIFF
--- a/app/views/application/subject-knowledge/index.html
+++ b/app/views/application/subject-knowledge/index.html
@@ -20,6 +20,8 @@
 {% block primary %}
   <p class="govuk-body">If you’re applying for secondary teacher training, describe your knowledge of the subjects you’d like to teach.</p>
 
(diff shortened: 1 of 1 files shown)


MaeveRoseveare:
[Investigate where else we could surface SKE information for candidates](https://trello.com/c/rc1jMLja/1075-investigate-where-else-we-could-surface-ske-information-for-candidates)

MylesJarvis:
Closing now we're not doing this.

PR #762, opened 2023-01-30
Upgrade prototype kit to v13

frankieroberto:
This upgrades the prototype for Apply to [version 13 of the prototype kit](https://prototype-kit.service.gov.uk/docs/whats-new).

Version 13 has some big and breaking changes, so there was a lot to update. I also took the opportunity to clean up and simplify a lot of the code, as that made the upgrade easier (although also harder).

Full list of changes:

* version 13 of prototype kit is now installed as a npm package (previously it wasn't a package dependency)
* the task list now uses the [task list plugin](https://github.com/x-govuk/task-list-component), which is a preview of an upcoming adding to the GOV.UK Design System
* the summary cards now use the code new added to govuk-frontend ([v4.5](https://github.com/alphagov/govuk-frontend/releases/tag/v4.5.0)) instead of custom code
* application data is no longer stored under a unique id within the session data (eg `data.application.1234`) and is instead just directly stored under `data.application`. This allows a lot of the routing code to be removed or simplified, and simplifies a lot of code (no more `applicationValue()` filter needed).
* all of the default session data has been removed. Instead there are now links in the footer which pre-fill applications and do other tasks.
* jQuery has been removed
* all the autocomplete code has been removed, for now, and replaced by their non-javascript fallbacks (either dropdowns or text inputs). Intention is to add this back in later using a plugin.
 * the (out-of-date) list of all providers, courses, subjects and so on has been removed, and replaced with a newer list of accredited providers for next cycle, and a hardcoded list of subjects/courses. This simplifies the code, and is probably good enough for testing, for now.
* the number of layouts has been reduced to just 2 (main and email), with the main one just adding header and footer, and not adding any grid layout or `<form>` tags. This does mean more duplication of code within each view, but I think that’s a bit less confusing and easier to maintain.
* `filters-with-app-context.js` has been removed - these no longer work with v13 of the prototype kit. The need for them has been removed through other refactoring and simplification.
* the javascript functions in `utils/index.js` have all been removed 
* the views and partials have been divided up into distinct folders representing different stages of the process - `application` (pre-submission), `equality-monitoring` (during submission), `survey` (post-submission interstitial)`, `dashboard` (post-submission, awaiting decisions), and `accepted` (once an offer has been accepted). `reference` contains the referee flow.

There’s a good chance that I’ve accidentally broken a few of the flows, but I think I’ve caught all of the main ones.

The footer contains these links which make it possible to quickly complete the application, and to advance to the next state where normally the user would be waiting for someone else to do something:

Screenshot 2023-02-16 at 17 30 50


frankieroberto:
@MaeveRoseveare @MylesJarvis if you have a spare moment, would either of you mind testing this using the review app: https://apply-tt-pr-762.herokuapp.com/

The main aim of this is to tidy up some tech debt and bring the prototype kit up to date so we can benefit from future changes. (If we start doing more design work using the production codebase it may prove to be less important, but let’s see...)

MaeveRoseveare (approved):
This looks good to me. I filled out the application form. Like the task list changes! Is there anything specific I should be looking for besides clicking through?

frankieroberto:
> This looks good to me. I filled out the application form. Like the task list changes! Is there anything specific I should be looking for besides clicking through?

Clicking through is fine. We can always fix anything we’ve missed later.

PR #764, opened 2023-02-17
Updated help text with email address

MaeveRoseveare:
Evidence from validation error logs shows that many candidates get errors when leaving the email address field blank when adding a reference (14,000+ errors as of Feb 2023).

We think this may be because they don't have the email address to hand. Adding "email address" to the intro screen may help by encouraging users to find the email address before starting.

[Trello card](https://trello.com/c/yY1Uy1Oe/1163-update-the-content-on-the-references-section-initial-page-to-make-it-clearer-that-youll-need-the-email-addresses-of-the-2-people)

Screenshot 2023-02-17 at 14 33 35


app/views/application/references/index.html
```diff
@@ -15,7 +15,7 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">{{ title }}</h1>
 
-      <p class="govuk-body">You need to give details of 2 people who can give a reference for you. They’ll be only be contacted if you accept an offer.</p>
+      <p class="govuk-body">You need to give the email address of 2 people who can give a reference for you. They’ll be only be contacted if you accept an offer.</p>
```
frankieroberto, line 18: When [copying this change to the production codebase](https://github.com/DFE-Digital/apply-for-teacher-training/pull/7952), I realised there’s a typo here ("be only be" 🤦) and also on production it says "an offer on a course" rather than "an offer" – not sure which of those is better?

```suggestion
      <p class="govuk-body">You need to give the email address of 2 people who can give a reference for you. They’ll only be contacted if you accept an offer on a course..</p>
```

frankieroberto (changes requested):

app/views/application/references/index.html
MaeveRoseveare: I think 'on a course' is better! Good spot!

app/views/application/references/index.html
```diff
@@ -15,7 +15,7 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">{{ title }}</h1>
 
-      <p class="govuk-body">You need to give details of 2 people who can give a reference for you. They’ll be only be contacted if you accept an offer.</p>
+      <p class="govuk-body">You need to give the email address of 2 people who can give a reference for you. They’ll only be contacted if you accept an offer on a course..</p>
```
MaeveRoseveare, line 18:
```suggestion
      <p class="govuk-body">You need to give the email address of 2 people who can give a reference for you. They’ll only be contacted if you accept an offer on a course.</p>
```

frankieroberto (approved):

PR #788, opened 2023-05-24
Add navigation bar and change header

frankieroberto:
This adds a navigation bar to the candidate interface, even though it only has 1 item in it for now (plus "sign out"). This is to enable further items to be added when continuous applications is launched.

🗂️ [Trello card](https://trello.com/c/wFS59p9L/1502-design-the-candidate-navigation-bar)

Changes:

* No longer show email address in black header bar
* Move sign out link from black header bar to grey navigation bar, aligned right
* Add "Your application(s)" as the first (and only) navigation item, and show it as selected.

Note: For now, the capitalisation of "Your application(s)" should follow the convention used on the service, which is:

* singular when not yet submitted
* singular if submitted but there was only 1 course chocie
* plural if submitted and there were 2+ course choices
* singular when an offer has been accepted

### Before

Screenshot 2023-05-24 at 13 57 42

### After

Screenshot 2023-05-24 at 15 23 01

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -5,6 +5,7 @@
 @import "components/section";
 @import "components/phase-banner";
 @import "components/header";
+@import "components/secondary-nav";
 
 .govuk-body {
   word-wrap: break-word;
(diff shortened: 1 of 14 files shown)


JR-G (approved):
🧑‍🎨

MylesJarvis (approved):

PR #810, opened 2023-07-07
Fixed 'withdrawal' and 'remove' link on application page

MaeveRoseveare:
I noticed on the prototype when an application is in "offer received' status, the link says 'Remove', but it should say 'Withdraw"
Small fix to update that.

**BEFORE:**
Screenshot 2023-07-07 at 09 47 13

**AFTER:**
Screenshot 2023-07-07 at 09 42 00


Kateharries (approved):
Looks great

PR #811, opened 2023-07-07
Match content on prototype to production

MaeveRoseveare:
We designed this page immediately on production but did not update the prototype. Small change to match the H1 and navigation tab content so people don't get confused.

This page shows when a candidate has accepted an offer, but if still waiting for their conditions to be met.

**BEFORE:**

Screenshot 2023-07-07 at 12 12 50

**AFTER:**

Screenshot 2023-07-07 at 12 12 18


Kateharries (approved):
Looks good :)

PR #814, opened 2023-07-12
Add designs for editing information after submission in Apply prototype

lsalivonaite-transform:
https://trello.com/c/zgdjm6SQ/340-apply-add-designs-for-editing-information-after-submission-in-apply-prototype

Also, fixed the 'complete/incomplete' tags

### DIFF
--- a/app/views/details/index.html
+++ b/app/views/details/index.html
@@ -25,7 +25,7 @@
         })}}
   {% else %}
    {{govukTag({
-          text: "Incompleted",
+          text: "Incomplete",
           classes: "govuk-tag--grey"
         })}}
   {% endif %}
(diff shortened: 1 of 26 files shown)


app/views/details/index.html
```diff
@@ -4,20 +4,34 @@
...
+{% endmacro %} #}
+
+{% macro tagTextForSection(completed=false) %}
+  {% if completed %}
+    {{govukTag({
+          text: "Completed"
+        })}}
+  {% else %}
+   {{govukTag({
+          text: "Incompleted",
```
MaeveRoseveare, line 28:
```suggestion
          text: "Incomplete",
```

MaeveRoseveare (changes requested):
Looks grand! I just noticed 2 things that I think we should remove.

1. On references (I know we said we shouldn't but now I think we should, SORRY)
Can we remove the grey button when they can't edit the section:

Screenshot 2023-07-13 at 11 27 54

2. On 'Other qualifications' - same thing can we remove the grey button when they can't edit:

Screenshot 2023-07-13 at 11 28 32

MaeveRoseveare (approved):

PR #817, opened 2023-07-31
Remove links from footer

MaeveRoseveare:
Small fix to remove redundant link from the footer of the prototype as it was confusing people.

Also changed content from 'Manage prototype' to 'Manage this prototype' because it sounded like we're talking about the Manage service.

**Before:**
Screenshot 2023-07-31 at 16 31 18

**After:**

Screenshot 2023-07-31 at 16 31 07

### DIFF
--- a/app/views/layouts/main.html
+++ b/app/views/layouts/main.html
@@ -72,7 +72,7 @@
       items: [
         {
           href: "/manage-prototype",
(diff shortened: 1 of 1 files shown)


lsalivonaite-transform (approved):
All good

PR #823, opened 2023-08-16
other nationalities path and fixes

lsalivonaite-transform:
Added back the questions for the other nationality branching, that will be needed for the 'visa sponsorship blocking pattern' later. 
And added redirects to review pages of every 'your details' section if the section is complete, as currently we need to go through all questions to get to the review pages.


app/views/details/personal-information/immigration-status.html
```diff
@@ -7,62 +7,29 @@
...
-      },
-      hint: {
-        text: "For example, “I have permanent residence”."
-      },
-      maxwords: 200,
-      rows: 5
-    }) }}
-  {% endif %}
+{% block content %}
+
```
MaeveRoseveare, line 15: The width seems very long on these pages, think we're missing the two-thirds tags:
```
{% block content %}
  <div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds">
...........
..........
.........
</div>
</div>

{% endblock %}
```

app/views/details/personal-information/immigration.html
```diff
@@ -18,24 +18,28 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
   <p class="govuk-body">You’ll need the right to work or study in the UK for the duration of your course.</p>
```
MaeveRoseveare, line 18: Same for this page, on line 15 we need those two-thirds tags.
```
{% block content %}
  <div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds">
...........
..........
.........
</div>
</div>

{% endblock %}
```

MaeveRoseveare (approved):
Looks great! And the pages navigation is so much better! I just left 2 comments about the width of the 2 new pages. But you can look on existing pages and steal from there.

PR #834, opened 2023-09-18
Interruption module

lsalivonaite-transform:
Added an interruption module page, to interrupt candidates before they submit their applications with advice that they need to improve their application.

Added 2 examples for personal statement and degree, which work in a way, but it needs improvement. Character reference example is not being done in the prototype.

Interruption-degree png
Interruption-personal-statement
Interruption-reference png


app/views/applications/interruption-module.html
```diff
@@ -0,0 +1,59 @@
...
+    value: "yes",
+    text: "Submit application",
+    classes: "govuk-button--inverse"
+  }) }}
+  {% if data.personalStatement.length < 500 %}
+  <a class="govuk-link" href="/details/personal-statement">Edit your personal statement</a>
+ {% elif degree.G3CL4.grade == "Third-class honours"%}
+   <a class="govuk-link" href="/applications">Save as a draft</a>
+  {% else%}
+   <a class="govuk-link" href="/applications">Add references</a>
```
MaeveRoseveare, line 52: Should this link to the references section? - /details/references

MaeveRoseveare (approved):

app/views/applications/interruption-module.html
lsalivonaite-transform: Yes

Kateharries (approved):
Looks great

PR #874, opened 2024-09-04
Updating candidate withdrawal reasons

PeteWilliams:
We have some upcoming work to update the reasons providers can select when rejecting an application.
This PR is to update the prototype to reflect the current journey on live, as it was quite different.

### Before
image

### After
image

### Live
image

### DIFF
--- a/app/filters.js
+++ b/app/filters.js
@@ -1,5 +1,9 @@
 const govukPrototypeKit = require('govuk-prototype-kit')
+const { prototypeFilters } = require('@x-govuk/govuk-prototype-filters')
+
 const addFilter = govukPrototypeKit.views.addFilter
+
+addFilter(prototypeFilters)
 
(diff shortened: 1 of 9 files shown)


PeteWilliams:
[Update prototype to reflect current live journey rejection reasons journeys](https://trello.com/c/1PEHPJlH/2257-update-prototype-to-reflect-current-live-journey-rejection-reasons-journeys)

simonwhatley:
@PeteWilliams, looking at the success banner in your _after_ screenshots, typically in BAT, the banner is full width as the content within is automatically two-thirds width.

PeteWilliams:
> @PeteWilliams, looking at the success banner in your _after_ screenshots, typically in BAT, the banner is full width as the content within is automatically two-thirds width.

Thanks, that's now updated.

simonwhatley (approved):

